PR #3, opened 2024-07-07
re-implement code using golang

kazzmir:
use golang and ebiten


kazzmir:
so far this just provides an lbx reader and xmi->midi converter

PR #153, opened 2024-12-26
city: use predefined city names

msom:
Hi @kazzmir

Great work on the Master of Magic clone!

This PR introduces a small enhancement: it suggests city names from the `cityname.lbx` file when creating a new settlement.

The name suggestions are based on the race of the settlers and whether the name has already been taken:
- Find the next name for a race not already taken.
- If all names for a particular race are already assigned, the next free name from any race is selected.
- If all predefined names are taken, a fallback suggestion of the format "City {number}" is provided.
- In contrast to the description in the [Master of Magic Wiki](https://masterofmagic.fandom.com/wiki/Town#Settlement_Names), no randomization is performed.

It seems that `cityname.lbx` contains an array of fixed-size strings, with both the size and the number of entries indicated by a small header. Comparing this structure with the [Master of Magic Wiki](https://masterofmagic.fandom.com/wiki/Town#Settlement_Names), it appears these strings are evenly distributed among the races, ordered alphabetically.

Let me know if you have any questions or need further adjustments.

### DIFF
--- a/game/magic/game/game.go
+++ b/game/magic/game/game.go
@@ -7,7 +7,9 @@ import (
     "log"
     "math"
     "fmt"
+    "regexp"
     "strings"
+    "strconv"
     "slices"
 
     "github.com/kazzmir/master-of-magic/game/magic/setup"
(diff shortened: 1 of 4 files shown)


kazzmir:
Thanks for contributing!

This PR is very good. I was thinking of doing it myself at some point. As a final fallback, "City N+1" is ok, but I was thinking that a first fallback can be one of the existing city names for a particular race with a number after it. For example suppose for race X the following city names are in cityname.lbx

```
A
B
C
```
Then if there are 3 cities in the game already with those names, the suggestion would be `A 2` or `B 2`. And if there are so many cities that `A 2`, `B 2`, etc are taken, then the next suggestion would be `A 3`.

Otherwise it seems slightly strange to suggest a klackon city name for a new barbarian city in case all the barbarian names are already taken.

BTW are you in the discord channel for master of magic?

game/magic/game/game.go
```diff
@@ -596,6 +596,61 @@ func (game *Game) FindValidCityLocationOnContinent(x int, y int) (int, int) {
...
+        case data.RaceTroll: raceIndex = 13
+    }
+
+    var existingNames []string
+    for _, city := range allCities {
+        existingNames = append(existingNames, city.Name)
+    }
+
+    entriesPerRace := len(predefinedNames) / 14
+    for i := 0; i < entriesPerRace; i++ {
```
kazzmir, line 638: It might be nice if this was randomized, which could be done with
```
for _, i := range rand.Perm(entriesPerRace) {
  ...
}
```

kazzmir:
Looks good! I'll merge this later today

msom:
Nice!

Although I joined this channel, I'm not much of a discord user...

game/magic/game/game.go
```diff
@@ -596,6 +598,75 @@ func (game *Game) FindValidCityLocationOnContinent(x int, y int) (int, int) {
...
+        if !slices.Contains(existingNames, name) {
+            return name
+        }
+    }
+    
+    // try to find a used city name without any number
+    expression := regexp.MustCompile(`(.*) (\d+)$`)
+    var matches []string
+    for _, i := range rand.Perm(entriesPerRace) {
+        name := predefinedNames[entriesPerRace * raceIndex + i]
```
kazzmir, line 652: I think theres a bug here in that the regexp shouldn't match the predefined name, since the predefined name will always just be `somecity` and not `somecity 2`. The regexp should match over all the existingNames instead. I'll just implement a simple loop instead.

kazzmir:
github won't let me edit your pr directly so I made this one https://github.com/kazzmir/master-of-magic/pull/154

PR #154, opened 2024-12-26
city: use predefined city names (#153)

kazzmir:
* city: use predefined city names

* Adjust city name fallback


kazzmir:
I factored out the part of the suggestion code that actually chooses a name so that I could write a test case that does not depend on the lbx file existing.

PR #160, opened 2024-12-29
Items

msom:
Hi @kazzmir 

I wondered how `itemdata.lbx` is structured and fell somehow down a rabbit hole. Except the very last byte of each item (seems to be some sort of flag), I think if got everything with the help of the wiki [1](https://masterofmagic.fandom.com/wiki/List_of_Conditionally-Available_Items), [2](https://masterofmagic.fandom.com/wiki/List_of_Forbidden_Items), [3](https://masterofmagic.fandom.com/wiki/List_of_Generally-Available_Items).

I'm not sure how to proceed at this point:
- The existing `Artifact` is missing some fields: Abilities, Mana Cost, Slot, Requirements. I've put it an own struct `Item` for now.
- Where to put the function that reads the predefined items. I've put it in `artifacts/items.go` for now.
- At which point to read the existing items.
- How to integrate the spells.


.gitignore
```diff
@@ -8,5 +8,7 @@
 data/data/
 
 # Executables
-./magic
-./lbxdump
+lbxdump
+lbxviewer
+magic
```
kazzmir, line 13: For me I need to put `/` at the front to ensure they only ignore the binaries. Otherwise just `magic` by itself will ignore everything under `game/magic`. Do you see a different behavior?

msom: Yes, with MacOS and git `2.43.0`  `./magic` does not work, but `/magic` would.

game/magic/artifact/item.go
```diff
@@ -0,0 +1,286 @@
...
+
+	"github.com/kazzmir/master-of-magic/game/magic/data"
+	"github.com/kazzmir/master-of-magic/lib/lbx"
+)
+
+// TODO: use Artifact
+type Item struct {
+	Name string
+	Image int
+	Slot ArtifactSlot
```
kazzmir, line 16: I guess the slot isn't really needed since the type implies the slot. I'm not really sure why the lbx file has it.

kazzmir:
I'm going to move all the types and stuff from create-artifact.go into items.go, and then use Artifact instead of Item

kazzmir:
now you can do `lbxdump magic.zip itemdata` and it will dump out all the items in text form
```
Artifact: {Type:1 Image:0 Name:Black Asp Powers:[+1 Attack] Abilities:[?]}
Artifact: {Type:1 Image:1 Name:Sword of Mallana Powers:[+1 Attack +1 Defense] Abilities:[]}
Artifact: {Type:1 Image:3 Name:Sword of Stealth Powers:[+3 Attack +3 Defense] Abilities:[?]}
```

kazzmir:
Oh yea I didn't even notice the tabs.. I have `expandtab` in vi and `tabstop=4` so a tab looks like 4 spaces to me. You probably noticed that I don't use gofmt.

kazzmir:
I think this PR is good for now. For the future, ReadArtifacts() can be called when creating treasure for a lair/cave when a random item is chosen.

Also I will probably make some simple gui using ebitenui that shows all the artifacts with their various properties.

msom:
Nice!

PR #166, opened 2024-12-31
artifact: add autogenerated names

msom:
Adds autogeneration of artifact names.

Names are generated as in the original game based on the selected powers, for example "+3 Orb of Power". Names change when changing powers and/or (misc) icons. Once a name has been changed manually, autogeneration is disabled.

Names of abilities and spells are not implemented yet.

I also added more `XYZBonus` functions, but I'm not sure if this fits your plans (especially regarding magic ranged attack).

### DIFF
--- a/game/magic/artifact/item.go
+++ b/game/magic/artifact/item.go
@@ -342,6 +342,15 @@ func (artifact *Artifact) RangedAttackBonus() int {
     }
 }
 
+func (artifact *Artifact) MagicAttackBonus() int {
+    switch artifact.Type {
+        case ArtifactTypeWand, ArtifactTypeStaff, ArtifactTypeMisc:
+            return addPowers[*PowerAttack](artifact.Powers)
+        default:
+            return 0
+    }
+}
+
 func (artifact *Artifact) DefenseBonus() int {
     base := addPowers[*PowerDefense](artifact.Powers)
     switch artifact.Type {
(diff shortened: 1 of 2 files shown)


kazzmir:
This is good. The only minor issue is when choosing a Chain or Plate item, the name is 'Chain of Defense' by default because a Chain item automatically has a DefenseBonus of 1, and Plate is 2.

image

I could see various ways of fixing this (invoke some method other than artifact.DefenseBonus(), or pass an argument into DefenseBonus() to ignore the default non-power bonuses), but those fixes seem inelegant compared to the impact of the issue.

kazzmir:
Well actually I think it might be ok to add a method `HasDefensePower` that returns true if the item has any powers of type `PowerDefense`. Similar methods can be added for the other powers, but can be elided until they are needed.

game/magic/artifact/item.go
```diff
@@ -354,10 +363,26 @@ func (artifact *Artifact) DefenseBonus() int {
     return base
 }
 
+func (artifact *Artifact) ToHitBonus() int {
```
kazzmir, line 366: I think these extra bonus methods are good, lets keep them

kazzmir:
I added most of the other HasXYZPower() functions, at least the ones used in the switch statement for the postfix name. I left melee/range/magic attack stuff out for now, but if they are useful they can be added later.

kazzmir (approved):

PR #169, opened 2025-01-02
artfifact: use power definitions from itempow.lbx

msom:
Use power definitions from `itempow.lbx`.

Sorry for the big PR with many changes 😬

Not 100% sure about this. The benefit would be that it also allows to play `Caster of Magic` and others. One drawback might be that there seems to be at least one bug in the original `itempow.lbx` (+6 Defense has actually only +5 Defense) - this makes it impossible to use the definitions from `itempow.lbx` for the items in `itemdata.lbx` (without a workaround).

Changes:
- Read item powers from `itempow.lbx` instead of defining the in the code.
- Use a simple struct for powers (including the name).
- Only store the cost of the whole artifact, not each power. This allows to use the actual costs from `itemdata.lbx` no matter if this matches the sum of costs defined in `itempow.lbx` (not sure if this is a use case).
- Use a cost threshold for decision whetever to show power in `Enchant Item` or `Create Artifact`. At this point, `200 Mana` seems to be the limit, but this requires revalidation when implementing the abilities (which I thought of doing next).

### DIFF
--- a/game/magic/artifact/create-artifact.go
+++ b/game/magic/artifact/create-artifact.go
@@ -6,7 +6,9 @@ import (
     "strings"
     "image/color"
     "log"
+    "bytes"
 
+    "github.com/kazzmir/master-of-magic/lib/set"
     uilib "github.com/kazzmir/master-of-magic/game/magic/ui"
     "github.com/kazzmir/master-of-magic/game/magic/util"
     "github.com/kazzmir/master-of-magic/game/magic/data"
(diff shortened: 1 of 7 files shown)


kazzmir:
For the +6 Defense bug, how about if the code does `if name == "+6 Defense" { amount = 6 }`.

game/magic/artifact/create-artifact.go
```diff
@@ -71,9 +299,44 @@ func getName(artifact *Artifact, customName string) string {
...
+            powerCost += costs[power]
+        }
+    }
+
+    // jewelry costs are 2x
+    if artifact.Type == ArtifactTypeMisc {
+        powerCost *= 2
+    }
+
+    // TODO: Artificer only pay half the full
```
kazzmir, line 332: I'm guessing the way to know if Artificer is used is a boolean flag can be passed in from ShowCreateArtifactScreen(), and passed all the way into this method as an argument.

kazzmir (approved):

PR #170, opened 2025-01-03
Add hire mercenaries event

msom:
This PR adds the hire mercanries event:

Bildschirmfoto 2025-01-03 um 06 08 27

Event generation is following the description [in the wiki](https://masterofmagic.fandom.com/wiki/Mercenaries):
- does not trigger if
  - early in the game
  - the player is being banished (no fortress city)
- generates the event with a chance of `max(10, 1 + fame / 20)`%
- selects a random unit from all units except
  - fantastic or hero
  - settlers
  - ships
  - races from an absent plane
- select a random number of units between 1 and 3
- selects random experience points of 20 (level 1) / 60 (level2) / 120 (level 3)
- costs `#units * production cost * (level + 3) / 2 [* 0.5 if charismatic]`
- puts the units in the fortress city if hired

Random unit selection is done slighty different: It first selects a list of possible candidates to choose from instead of trying 20 times randomly a unit from all units and fail if the unit does not match the criteria.

Presence on a plane is also done slighty simpler: Presence is given only if a player has a city on a plane instead of a city or unit or tower of wizardry.

### DIFF
--- a/game/magic/game/game.go
+++ b/game/magic/game/game.go
@@ -1919,6 +1919,10 @@ func (game *Game) doHireHero(yield coroutine.YieldFunc, cost int, hero *herolib.
 /* show the hire mercenaries popup, and if the user clicks 'hire' then add the units to the player's list of units
  */
 func (game *Game) doHireMercenaries(yield coroutine.YieldFunc, cost int, units []*units.OverworldUnit, player *playerlib.Player) {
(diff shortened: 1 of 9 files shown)


kazzmir:
wow nice!!

kazzmir (approved):

PR #172, opened 2025-01-03
Add merchant event

msom:
Adds the merchant event:

Bildschirmfoto 2025-01-03 um 18 24 53

All predefined items are read from the `itemdata.lbx` and added to a pool, where they can be taken out by the merchant (or later for lair treasure).

Also adds requirements (number of books in certain magic types) to artifacts and reads them for the predefined items from `itemdata.lbx`

Event generations follows the description in the [wiki](https://masterofmagic.fandom.com/wiki/Merchant):
- generates the event with a chances of `max(10, 2 + player.Fame / 25 [*2 if famous])`
- cost is `artifactCost [/ 2 if charismatic]`
- an item is skipped, if the requirements are not met (the merchant has 12 books in every type of magic)

I further moved the part of drawing an artifact also used in the vault to a new `artifactview` folder.

@kazzmir What do you think?

### DIFF
--- a/game/magic/artifact/item.go
+++ b/game/magic/artifact/item.go
@@ -125,6 +125,11 @@ type Power struct {
     Spell spellbook.Spell
 }
 
+type Requirement struct {
+    MagicType data.MagicType
+    Amount int
+}
+
 type Artifact struct {
     Type ArtifactType
     Image int
@@ -132,6 +137,7 @@ type Artifact struct {
     Cost int
     Powers []Power
     Abilities []data.Ability
+    Requirements []Requirement
 }
 
 func (artifact *Artifact) HasAbility(ability data.AbilityType) bool {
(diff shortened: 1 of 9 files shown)


kazzmir:
Ok this is good. Usually I create the `xyzview` directories to get around a circular import, but in this case `artifactview` doesn't depend on anything that in turn depends on `artifactview`. Could we keep the code in `artifact` for now? `RenderArtifactBox` can be in its own file, still with the `view.go` name.

kazzmir:
image

btw the merchant screen seems to get rendered starting at 0,0. Is that how its supposed to look? Or should there be some margin on the left/top. Its fine either way.

kazzmir:
image

one more minor issue, I think 'spell save' should render as '-2 spell save'

msom:
Also adds the artifact name to the hero popup:
Bildschirmfoto 2025-01-04 um 06 54 43

kazzmir (approved):

PR #181, opened 2025-01-05
Add hero level up message

msom:
Adds the hero level up popup:

Bildschirmfoto 2025-01-05 um 11 35 58

The overworld scenario 26 contains two heros:
- Gunther will gain the next level at the start of the turn
- Raywind will gain the next level if combat against the spearman is won

Does not render the ability improvements (they seem not to be implemented yet).

### DIFF
--- a/game/magic/hero/hero.go
+++ b/game/magic/hero/hero.go
@@ -972,8 +972,6 @@ func (hero *Hero) GetBaseHitPoints() int {
 }
 
 func (hero *Hero) getBaseHitPointsProgression(level units.HeroExperienceLevel) int {
-    // FIXME: shouldn't this be 10% starting from captain?
-    //        see https://masterofmagic.fandom.com/wiki/Experience_Level#Attribute_Improvement_Table_-_Heroes
     switch level {
         case units.ExperienceHero: return 0
         case units.ExperienceMyrmidon: return 1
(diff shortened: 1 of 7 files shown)


game/magic/hero/hero.go
```diff
@@ -964,22 +967,57 @@ func (hero *Hero) GetHitPoints() int {
...
 
 func (hero *Hero) GetBaseHitPoints() int {
-    base := hero.Unit.GetBaseHitPoints()
-
     level := hero.GetExperienceLevel()
+    return hero.Unit.GetBaseHitPoints() + hero.getBaseHitPointsProgression(level)
+}
+
+func (hero *Hero) getBaseHitPointsProgression(level units.HeroExperienceLevel) int {
+    // FIXME: shouldn't this be 10% starting from captain?
```
kazzmir, line 975: thats for the 'to hit' modifier, which is `GetToHitMelee()`

kazzmir: this method is the hearts

kazzmir:
I was surprised to see that dos MoM uses the text 'xyz has made a level'. The verb 'made' is pretty awkward there. I think 'xyz has gained a level' would be better, but we can keep 'made' since the original game had it.

game/magic/hero/hero.go
msom: You're ight, I goth them mixed up!

kazzmir (approved):

PR #196, opened 2025-01-12
lbxdump: dump terrain tiles by index

msom:
Currently, finding the correct terrain image for a tile is challenging because images from an animation group increment the counter in the filename.

Using `ReadTerrainData` allows the dumped files to be named separately by their index and animation index (although dumping now relies on the tile definitions in terrain and not only on the file itself anymore).

I suggest showing both the index in decimal (4 digits) as hex. This way the files are ordered nicely and the image for a tile index can be found easily.

Example: `image_0201_0x0c9_0`


kazzmir:
Ok seems fine. Do you use the terrain viewer utility?

msom:
Oh, I wasn't aware of that. Nice! Never mind this PR then!

PR #200, opened 2025-01-14
map: improve tile resolution

msom:
I think there are some issues with the tile resolution:
- Center tiles need to be evaluated too, not only the surrounding
- Some tiles care not only what their neighbors are, but oftentimes what they are **not**
- Not all tiles care about **all** neighbors
- Some tiles allow neighbors to be different types (or require to exclude different types)

For example:
- Grassland tiles only care about the center (is it grassland or not)
- Hill tiles only care if there is or is not a hill tile top/left/bottom/right
- Shore tiles only care if its either ocean or shore or neither ocean or shore
- Some shore tiles like bays do not care about the neighbors in the corner of the bay

With this, a single pass resolves all the tiles.

By using this, shore generation looks more like the original and does not create artifacts:
Old:
Bildschirmfoto 2025-01-14 um 18 23 40
New:
Bildschirmfoto 2025-01-14 um 18 27 29

Hills (new):
Bildschirmfoto 2025-01-14 um 18 25 10

Mountains (new):
Bildschirmfoto 2025-01-14 um 18 25 29

Desert (new):
Bildschirmfoto 2025-01-14 um 18 25 47

Tundra (new):
Bildschirmfoto 2025-01-14 um 18 26 07


msom:
Also updates the terrainviewer:
Bildschirmfoto 2025-01-14 um 18 48 15

kazzmir:
awesome! in the terrain viewer I see there are a couple of neighbors marks as 'unknown'. does 'unknown' mean basically that any tile can be attached there?

image

msom:
Yes, there is no condition in the map so it renders unknown.

PR #201, opened 2025-01-14
map: make random terrain placement appear less arbitrary

msom:
Make the random terrain placement a bit less arbitrary looking by:
- placing tundra only near the poles
- placing desert only near the center
- use some weights
- generate a bit less land in general
- generate a bit less land at the poles

Bildschirmfoto 2025-01-14 um 21 09 20

### DIFF
--- a/game/magic/terrain/map.go
+++ b/game/magic/terrain/map.go
@@ -60,7 +60,7 @@ func (map_ *Map) FloodWalk(x int, y int, f FloodFunc){
     walk(x, y)
 }
 
-func (map_ *Map) FindContinents(plane data.Plane) []Continent {
+func (map_ *Map) FindContinents() []Continent {
 
     seen := makeCells(map_.Rows(), map_.Columns())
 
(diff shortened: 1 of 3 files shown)


game/magic/terrain/map.go
```diff
@@ -340,32 +377,29 @@ func (map_ *Map) PlaceRandomTerrainTiles(plane data.Plane){
...
         for i := 0; i < int(math.Sqrt(float64(continent.Size()))) / 8; i++ {
             point := chooseRandomElement(continent)
 
             var use int
-            switch rand.IntN(4) {
+            switch rand.IntN(3) {
                 case 0: use = TileSorceryLake.Index(plane)
                 case 1: use = TileNatureForest.Index(plane)
                 case 2: use = TileChaosVolcano.Index(plane)
-                case 3: use = TileVolcano.Index(plane)
```
kazzmir, line 368: Are volcano tiles still generated somewhere?

kazzmir:
On the image you pasted I see some places where the land tiles discontinually stops and immediately becomes ocean. Is that due to some wrapping issue?

kazzmir:
Actually I don't see a discontinuity when I generate land on my machine, so I guess its fine.

What do you think about generating rivers/lakes?

game/magic/terrain/map.go
msom: This was only for testing, I guess volcano's only appear through "raise volcano" https://masterofmagic.fandom.com/wiki/Volcano#Creating_Volcanoes

msom:
Yes, I think it only looks like this in this picture.

Lakes and rivers might need further condition checking (two adjacent lakes would make shore; there are not tiles for every river scenario) - maybe better in a future PR?

kazzmir:
ok looks good, ill merge in a bit

PR #205, opened 2025-01-15
artifact: consider artificer ability when creating and destroying artifacts

msom:


kazzmir:
What about runemaster? I think that takes off an additional 25% from the artifact cost

msom:
I actually thought about adding it but somehow assumed it would be better to implement this near spell casting since these 25% are for all arcane spells...?

kazzmir:
Ok you are right

kazzmir:
Actually, if you have runemaster does that get reflected in the cost shown in the create artifact screen? I'll test it out in dos mom

kazzmir (approved):

PR #212, opened 2025-01-17
map: add rivers

msom:
Adds rivers:
Bildschirmfoto 2025-01-17 um 13 16 41

Creating rivers is done by performing random walks:
- each walk ends at a shore or lake
- each walk cannot come adjacent to another river as this will render as loops
- each walk must be at least 2 tiles (1 tile rivers are not rendered correctly)
- each walk must be fully resolvable (since not all shore/river constellation are present with a tile).

Also contains some code cleanup.


kazzmir:
Ok looks cool. Should the river be able to split into branches?

Can we use roughly this same logic for hills and mountain ranges?

msom:
I couldn't figure out how to avoid loops with branches (when walking from inland to shore).

Not sure how this turns out with hills and mountains, we could try I guess...

kazzmir:
ok ill merge this for now, and we can work on branches/hills/mountains later

util/mapeditor/map_test.go
```diff
@@ -1,22 +0,0 @@
-package main
```
kazzmir, line 1: can we keep this? I thought it might be useful to benchmark how fast map generation is.

msom: i moved it to the map_test in the map package

msom: https://github.com/kazzmir/master-of-magic/pull/212/files/00de0eb3bf48717a0617e20ae204aaaf467e9da3#diff-9accd79cf766bce1148df2448d882eabf188c8be047b85f4b5522419865bc579

kazzmir: oh ok :+1:

kazzmir:
Ok just so I understand, the river generation chooses some land tile and then does a random walk until it hits an end tile (ocean/shore), while avoiding other river tiles.

msom:
> Ok just so I understand, the river generation chooses some land tile and then does a random walk until it hits an end tile (ocean/shore), while avoiding other river tiles.

Yes

kazzmir:
One possible optimization is that each continent can have rivers placed on it in a separate goroutine. We can do it later if necessary.

kazzmir (approved):

PR #213, opened 2025-01-17
map: create nodes from terrain

msom:
Creates `maplib` nodes from `terrain` nodes.

Also use the automatically created nodes in the overworld tests instead of creating new nodes.

Also adds a small fix to avoid placing cities on nodes.


test/overworld/main.go
```diff
@@ -40,6 +41,25 @@ type Engine struct {
     Console *console.Console
 }
 
+type NodeInfo struct {
+    X int
+    Y int
+    Node *maplib.ExtraMagicNode
+}
+
+func findNodes(mapObject *maplib.Map) *map[terrain.TerrainType][]NodeInfo {
```
kazzmir, line 50: why does this return a pointer?

msom: you're right, that's not necessary

test/overworld/main.go
```diff
@@ -594,10 +610,19 @@ func createScenario8(cache *lbx.LbxCache) *gamelib.Game {
 
     drake := player.AddUnit(units.MakeOverworldUnitFromUnit(units.GreatDrake, x + 1, y + 1, data.PlaneArcanus, wizard.Banner, nil))
 
-    for i := 0; i < 1; i++ {
-        fireElemental := player.AddUnit(units.MakeOverworldUnitFromUnit(units.FireElemental, x + 1, y + 1, data.PlaneArcanus, wizard.Banner, nil))
-        _ = fireElemental
-    }
+    nodes := findNodes(game.CurrentMap())
```
kazzmir, line 613: I generally agree with having the map create nodes, but when this test case is run the nodes appear in random places on the map so its not super easy to have a unit walk enter one of the nodes.

Can we still have the CreateNode() function just for testing purposes?

msom: that's why I put additional fire elementals near the other types of nodes

msom: but sure, we can leave the createnode

kazzmir (approved):

PR #220, opened 2025-01-18
spell: add change terrain

msom:
Adds change terrain spell implementation.

There are some slight differences to the original:
- camera is centered to the changing tile before animation
- instead of showing a red box with "That terrain may not be changed", invalid tiles are just not selectable

@kazzmir what do you think?

I was also wondering if `ScreenToTile` should actually return wrapped x?

Also update the TODO.

### DIFF
--- a/game/magic/game/cast.go
+++ b/game/magic/game/cast.go
@@ -18,6 +18,7 @@ import (
     "github.com/kazzmir/master-of-magic/game/magic/summon"
     "github.com/kazzmir/master-of-magic/game/magic/util"
     "github.com/kazzmir/master-of-magic/game/magic/audio"
+    "github.com/kazzmir/master-of-magic/game/magic/terrain"
     // "github.com/kazzmir/master-of-magic/game/magic/data"
 
     "github.com/hajimehoshi/ebiten/v2"
@@ -30,6 +31,7 @@ const (
     LocationTypeEnemyCity
     LocationTypeFriendlyUnit
     LocationTypeEnemyUnit
+    LocationTypeChangeTerrain
 )
 
 func (game *Game) doCastSpell(yield coroutine.YieldFunc, player *playerlib.Player, spell spellbook.Spell) {
@@ -76,7 +78,7 @@ func (game *Game) doCastSpell(yield coroutine.YieldFunc, player *playerlib.Playe
             cityview.PlayEnchantmentSound(game.Cache)
             game.showCityEnchantment(yield, chosenCity, player, spell.Name)
         case "Change Terrain":
-            tileX, tileY, cancel := game.selectLocationForSpell(yield, spell, player, LocationTypeAny)
+            tileX, tileY, cancel := game.selectLocationForSpell(yield, spell, player, LocationTypeChangeTerrain)
 
             if cancel {
                 return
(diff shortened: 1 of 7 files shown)


game/magic/game/cast.go
```diff
@@ -75,6 +77,14 @@ func (game *Game) doCastSpell(yield coroutine.YieldFunc, player *playerlib.Playe
             yield()
             cityview.PlayEnchantmentSound(game.Cache)
             game.showCityEnchantment(yield, chosenCity, player, spell.Name)
+        case "Change Terrain":
+            tileX, tileY, cancel := game.selectLocationForSpell(yield, spell, player, LocationTypeChangeTerrain)
```
kazzmir, line 81: couldn't this just use `LocationTypeAny`? My plan was to have just a few location types. But changing terrain shouldn't allow the user to pick an ocean tile, so should this be something like `LocationTypeLand` ?

game/magic/game/cast.go
```diff
@@ -172,6 +182,7 @@ func (game *Game) selectLocationForSpell(yield coroutine.YieldFunc, spell spellb
     var selectMessage string
 
     switch locationType {
+        case LocationTypeChangeTerrain: fallthrough
```
kazzmir, line 185: just in general I prefer to not use fallthrough, and instead combine cases like
```
case X, Y: ...
```
instead of
```
case X: fallthrough
case Y: ...
```

game/magic/game/cast.go
msom: also tundra

game/magic/game/cast.go
```diff
@@ -307,6 +318,18 @@ func (game *Game) selectLocationForSpell(yield coroutine.YieldFunc, spell spellb
                         if city != nil {
                             return tileX, tileY, false
                         }
+                    case LocationTypeChangeTerrain:
```
kazzmir, line 321: I see, yes lets name this `LocationTypeLand`, as that is more general than just changing terrain, so that other spells might use the same location type

kazzmir:
currently ScreenToTile() can't call WrapX on its X return value because it messes up how the camera moves when you right click on the overworld.

game/magic/game/cast.go
kazzmir: well actually I guess other land-targeting spells have their own constraints that don't really generalize. like 'Raise Volcano' can target grassland, forest, swamp, desert, tundra, but not hill, mountain, or volcano. So I guess having a specific location type for now is fine.

kazzmir:
currently choosing a tile allows the user to select tiles under fog. I would guess this shouldn't be allowed

kazzmir:
ill merge this in a few hours

game/magic/game/cast.go
```diff
@@ -307,6 +317,21 @@ func (game *Game) selectLocationForSpell(yield coroutine.YieldFunc, spell spellb
...
```
kazzmir, line 320: in a follow up pr im going to change these locationType things to just a `func (x int, y int) bool` so that if it returns true then the location is valid.

kazzmir:
actually I was able to have ScreenToTile() call WrapX() on its x return value, so that should make things easier going forward

kazzmir (approved):

PR #222, opened 2025-01-18
Spell help

msom:
Adds help to spell book:

Bildschirmfoto 2025-01-18 um 16 18 14

Works for casting spell book and when choosing a new spell to learn, not however when viewing the spell book via apprentice. It looks like this would need a bigger refactor to work?

### DIFF
--- a/game/magic/spellbook/spell.go
+++ b/game/magic/spellbook/spell.go
@@ -271,6 +271,18 @@ func ShowSpellBook(yield coroutine.YieldFunc, cache *lbx.LbxCache, allSpells Spe
         return
     }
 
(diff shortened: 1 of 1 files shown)


kazzmir:
nice! Yea for the apprentice view the issue is that the spellbook basically ignores left click/right click except for the left/right dog ear elements. I'll look into refactoring the code so that help can work in the apprentice view.

Also note that the name passed to help is not always right. It is annoying that the names of things mostly match up with the name in help, but not always. For example, in ` test/cast-spell` no help comes up if you right click on 'Floating Island'. The help entry for 'Floating Island' is likely spelled slightly differently. Probably we will have to add something like `spell.HelpName` that if non-empty will override the spell name when looking up the help item.

PR #233, opened 2025-01-20
new-game: allow going back

msom:
Allows going back the screens by pressing escape.

Not proud of https://github.com/kazzmir/master-of-magic/compare/master...refs/heads/start-screen#diff-775028b416df7bdc9c4cf5f3941c9807169b6987c30aab5531a0d138eed44672L263. It feels very hacky and against the principles of state machines. Maybe better integrate the settings screens into the state machine of new-wizard?

### DIFF
--- a/game/magic/main.go
+++ b/game/magic/main.go
@@ -72,18 +72,20 @@ func runIntro(yield coroutine.YieldFunc, game *MagicGame) {
     }
 }
 
-func runNewGame(yield coroutine.YieldFunc, game *MagicGame) setup.NewGameSettings {
+func runNewGame(yield coroutine.YieldFunc, game *MagicGame) (bool, setup.NewGameSettings) {
     newGame := setup.MakeNewGameScreen(game.Cache)
 
     game.Drawer = func(screen *ebiten.Image) {
         newGame.Draw(screen)
     }
 
-    for newGame.Update() == setup.NewGameStateRunning {
+    state := newGame.Update()
+    for state == setup.NewGameStateRunning {
         yield()
+        state = newGame.Update()
     }
 
-    return newGame.Settings
+    return state == setup.NewGameStateCancel, newGame.Settings
 }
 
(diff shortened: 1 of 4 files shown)


kazzmir:
ok ill review the logic and state machine stuff to see what can be done cleaner

game/magic/setup/new-game.go
```diff
@@ -271,6 +271,15 @@ func (newGameScreen *NewGameScreen) MakeUI() *uilib.UI {
                 element.Draw(element, screen)
             })
         },
+        HandleKeys: func(keys []ebiten.Key){
+            for _, key := range keys {
+                switch key {
+                    case ebiten.KeyEscape:
```
kazzmir, line 277: btw on my system I have to add the `ebiten.KeyCapslock` here because I switch the escape and capslock keys
```
/usr/bin/setxkbmap -option "caps:swapescape"
```
https://github.com/hajimehoshi/ebiten/issues/2974 is the bug I logged in ebiten for this, but it seems difficult for them to fix it.

msom: good to know, i was wondering why capslock is everywhere :)

kazzmir: probably I can add something like `if inputmanager.IsQuitKey(key)` so its more compact

kazzmir (approved):

game/magic/main.go
```diff
@@ -259,12 +263,24 @@ func runGame(yield coroutine.YieldFunc, game *MagicGame, dataPath string) error
...
-                // yield so that clicks from the menu don't bleed into the next part
-                yield()
-                settings := runNewGame(yield, game)
-                yield()
-                wizard := runNewWizard(yield, game)
+                var settings setup.NewGameSettings
+                var wizard setup.WizardCustom
+                restart := true
+                cancel := false
+                for restart && !cancel {
```
kazzmir, line 270: this part is ok, and its locallized to just this loop so not too bad

PR #237, opened 2025-01-22
lbxdump: save sounds as wav

msom:
Allows to dump the sound files as `.wav` files.


kazzmir:
Sure this is fine. Also note that util/vocdump dumps the voc files, but saving as wav is ok

kazzmir:
Maybe this could support a `-voc` and `-wav` to select the output format? Then we could just remove vocdump

msom:
Ah, didn't know `voc` file format. Sure, I'll add this option!

kazzmir (approved):

PR #241, opened 2025-01-24
cast: implement raise volcano

msom:
Implement [Raise Volcano](https://masterofmagic.fandom.com/wiki/Raise_Volcano) Spell.

Volcanoes:
- may be cast on desert, forest, swamp, frass, and tundra
- destroys any bonus
- may destroy buildings with a chance of 15%
- generate 1 power for the casting wizard
- have a chance of 2% of reverting back to mountain, and thereby a chance of 5% of generating a random mineral
- can be reverted also with change terrain

I also reused the `Melder` interace but renamed it to `Wizard`, not sure if that is a good idea.

I also refactored the spells that affect the map. They now all behave the same way: for example centering on the tile and resetting zoom. The common/duplicated code is now in `doCastOnMap`. Since some things need to happen mid-animation (for example changing terrain when the animation covers most of the tile), `doCastOnMap` accepts an `UpdateMapFunction`. I also moved the spell logic to the `doXYZ` for better clarity.

### DIFF
--- a/game/magic/maplib/map.go
+++ b/game/magic/maplib/map.go
@@ -26,7 +26,7 @@ type CityProvider interface {
     ContainsCity(x int, y int, plane data.Plane) bool
 }
 
-type Caster interface {
+type Wizard interface {
     GetBanner() data.BannerType
 }
 
@@ -331,7 +331,7 @@ type ExtraMagicNode struct {
     Zone []image.Point
 
     // if this node is melded, then this player receives the power
-    MeldingWizard Caster
+    MeldingWizard Wizard
     // true if melded by a guardian spirit, otherwise false if melded by a magic spirit
     GuardianSpiritMeld bool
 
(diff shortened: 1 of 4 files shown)


msom:
I also fixed the enchant road sound effect.

BTW: any idea why soundfx.lbx contains the same sound over and over?

kazzmir:
> I also fixed the enchant road sound effect.
> 
> BTW: any idea why soundfx.lbx contains the same sound over and over?

No idea! It's crazy right?

game/magic/maplib/map.go
```diff
@@ -26,7 +26,7 @@ type CityProvider interface {
     ContainsCity(x int, y int, plane data.Plane) bool
 }
 
-type Melder interface {
+type Wizard interface {
```
kazzmir, line 29: sure this is fine

kazzmir (approved):
works great!

PR #244, opened 2025-01-25
map: unify usage of terrain tile bonuses

msom:
Rely (only) on `FoodBonus`, `GoldBonus` and `ProductionBonus` of a tile.

Since some of these Bonuses rely on neighboring tiles in case of River Mouths, I've moved these definitions from `terrain` to `map`. This will also be needed when disabling the bonuses in case of corruption.

I've also moved the gold bonus logic previously in `city` to `GoldBonus` together with `IsRiverMouth` and `IsTouchingShore`. This also allows the surveyor to display the gold bonus and river shores.

TBH, the wiki is very confusing (maybe sometimes contradicting) regarding these bonuses and the original surveyor seems to report many errors. Although this implementation might be more correct, most of the informations of the surveyor per se are actually not useful as it's not clear how the bonuses apply (e.g. gold bonus only if city is built there etc.). However, I would leave it as it is since people will probably expect to behave similar to the original (or patched) surveyor...

Surveyor now displays "River mouths" and "Gold Bonus":
Bildschirmfoto 2025-01-25 um 11 12 36

### DIFF
--- a/game/magic/maplib/map.go
+++ b/game/magic/maplib/map.go
@@ -499,7 +499,6 @@ func (tile *FullTile) IsRiverMouth(mapObject *Map) bool {
 }
 
 func (tile *FullTile) IsTouchingShore(mapObject *Map) bool {
-    touchingShore := false
     for dx := -1; dx <= 1; dx++ {
         for dy := -1; dy <= 1; dy++ {
             if dx == 0 && dy == 0 {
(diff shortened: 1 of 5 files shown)


game/magic/maplib/map.go
```diff
@@ -419,10 +420,102 @@ type FullTile struct {
...
+    touchingShore := false
+    for dx := -1; dx <= 1; dx++ {
+        for dy := -1; dy <= 1; dy++ {
+            if dx == 0 && dy == 0 {
+                continue
+            }
+
+            tile := mapObject.GetTile(tile.X + dx, tile.Y + dy)
+            if tile.Tile.IsShore() {
+                touchingShore = true
```
kazzmir, line 511: This might as well be return true here

kazzmir:
Looks good overall

kazzmir (approved):

PR #245, opened 2025-01-25
cast: implement corruption

msom:
Adds corruption map extra and corruption spell.

I'm not 100% sure what the effect of corruption should be. The [wiki](https://masterofmagic.fandom.com/wiki/Corruption_(Spell)) is somewhat confusing, so I tested what the influence of corruption on existing towns in the (patched) original game is and I'm pretty sure it just disables all bonuses (= `BonusNone`) and all Food/Gold/Production bonus of tiles (=0).

This is also reflected in the surveyor:
Bildschirmfoto 2025-01-25 um 18 53 32


kazzmir (approved):
nice!

PR #253, opened 2025-01-26
spell: implement warp node

msom:
image


game/magic/maplib/map.go
```diff
@@ -361,16 +363,44 @@ func (node *ExtraMagicNode) DrawLayer2(screen *ebiten.Image, imageCache *util.Im
...
+
+    if node.Warped && node.MeldingWizard != nil {
+        shader, _ := imageCache.GetShader(shaders.ShaderWarp)
+
+        point:= node.Zone[0]
+
+        mask, _ := imageCache.GetImage("mapback.lbx", 93, 0)
+
+        // FIXME: is there a better way to render the shader on the screen?
+        tx, ty := options.GeoM.Element(0, 2), options.GeoM.Element(1, 2)
```
msom, line 383: @kazzmir Is there an easier way to apply a shader to the current screen than creating a sub imgage?

kazzmir: I think thats how its supposed to work with ebiten. Its not super performant, but as long as the number of warped nodes is somewhat low it should be ok. The one platform that has the most issues with these sorts of effects is the browser, so I'll test there a bit.

game/magic/maplib/map.go
```diff
@@ -358,19 +360,57 @@ func (node *ExtraMagicNode) DrawLayer2(screen *ebiten.Image, imageCache *util.Im
...
+        rect := image.Rect(int(tx), int(ty), int(tx) + tileWidth, int(ty) + tileHeight)
+        image := screen.SubImage(rect)
+
+        if image.Bounds().Dx() == tileWidth && image.Bounds().Dy() == tileHeight {
+            var options2 ebiten.DrawRectShaderOptions
+            options2.GeoM = options.GeoM
+            options2.GeoM.Translate(float64(point.X * tileWidth), float64(point.Y * tileHeight))
+            options2.Images[0] = ebiten.NewImageFromImage(image)
```
kazzmir, line 401: is NewImageFromImage needed here? I'm not sure

msom: If I rembember correctly it will else complain about using the same image data for shader input and output. NewImageFromImage seems to copy the pixels?

kazzmir: someone online just said that NewImageFromImage isn't needed. screen.SubImage(rect) always returns a *ebiten.Image so you can just do
```
image := screen.SubImage(rect).(*ebiten.Image)
```
and give image to `options2.Images[0]` directly

msom: This will give a `panic: buffered: Image.DrawTriangles: source images must be different from the receiver`

kazzmir: oh ok, its fine probably to just make a copy then

kazzmir:
The sparkles render in the wrong place when the zoom is changed

msom:
> The sparkles render in the wrong place when the zoom is changed

Yes (also in the master), I added a FIXME: https://github.com/kazzmir/master-of-magic/pull/253/files#diff-27552e5d7f552fea342f121f0c244f4b30a0de5b7dc48d6b46a72e9a54878ee9R372

msom:
There is also a bug where the sparkles are not rendered depending on the screen position but I haven't figured out what the exact conditions are, maybe something with wrap?

kazzmir:
oh yea I noticed that too. I'll take a look at whats wrong

msom:
A zone that does that is `[]image.Point len: 9, cap: 16, [{X: 0, Y: 0},{X: -2, Y: 2},{X: 2, Y: -2},{X: -2, Y: 1},{X: 0, Y: -2},{X: 1, Y: 2},{X: -1, Y: 1},{X: 1, Y: 1},{X: -2, Y: -2}]`

kazzmir:
I fixed the issue with the sparkles moving based on zoom. The issue is that the options.GeoM passed in has to be Concat()'d to the geom used to draw the sparkle. This is because the Scale() transformation in the options.GeoM has to be last so that the total order of geom operations is
```
translate(tileX, tileY)
translate(sparkleX, sparkleY)
scale(screen scale)
```
The way the sparkle was doing things before it was
```
translate(tileX, tileY)
scale(screen scale)
translate(sparkleX, sparkleY)
```

I think the sparkles don't show up sometimes is because the node is off the screen, so thet DrawLayer2 code decides not to draw it. I'll just have the draw always happen

kazzmir:
I think the sparkles always show up now regardless of where you move the map. I mostly just copied the logic from DrawLayer1

kazzmir:
The warp sprite probably needs the same Concat() fix.
Normal zoom
image

Zoomed out
image

kazzmir:
I think the issue is with tileWidth/tileHeight. They are always the same size but the warp image should take the current zoom scale into account. I think the bottom right corner can just be options.GeoM.Apply(tileWidth, tileHeight) or something

game/magic/maplib/map.go
kazzmir: Why did you use Element() here? Did you mean to use Apply() ? I think Element() returns a value from the inside the matrix itself, while Apply() applies to the matrix to a vector(x,y) and returns the updated x,y coordinates

msom: I wanted to get the actual translation from the options as I thought options has already been moved to the tile.

msom: Couldn't think of a better way to get the screen coordinates of the tile from options

kazzmir:
The warp code compute the rect where the shader should apply while accounting for zoom now, but I broke the mask a little bit. The logic now is

```
set up geom where the tile would be drawn normally. geom.Translate(pointX, pointY); geom.Concat(options.Geom)
x1, y1 = geom.Apply(0, 0) // get upper left hand on the screen
x2, y2 = geom.Apply(tileWidth, tileHeight) // get lower right on the screen
subimage = screen.SubImage(rect(x1, y1, x2, y2)) get real portion of screen that matches up to that area
geom.Reset(); geom.Translate(x1, y1) // move to where the screen portion should be redraw. notably the scale is not present here
DrawRectShader() // as normal
```
One of the issues while doing this is that the screen image can vary depending on the zoom level, which means the mask image sprite has to be the same size. Currently I just do 'mask = ebiten.NewImage(subimage.Bounds())' to get something to work, but the mask should really just be a scale of the mask sprite.

msom:
With the current effect, the mask hardly makes any difference anyway. Maybe remove it altogether?

kazzmir:
yea haha that might be the best thing for now. we can always try to add the mask in later

kazzmir:
seems ok in a browser too. btw if you want to test the game in a browser you can do
```
wasmserve ./test/overworld 34
```
then open localhost:8080

wasmserve comes from https://github.com/hajimehoshi/wasmserve

kazzmir (approved):

PR #254, opened 2025-01-26
spell: implement warp node

msom:
Like https://github.com/kazzmir/master-of-magic/pull/253 but with black sparkles instead of shader:

Bildschirmfoto 2025-01-26 um 20 00 52


msom:
Or with more transparent sparkles (lower node):
Bildschirmfoto 2025-01-26 um 20 02 31

PR #257, opened 2025-01-27
help: add autocorrection

msom:
Tries to autocorrect name when getting help.

### DIFF
--- a/lib/lbx/help.go
+++ b/lib/lbx/help.go
@@ -36,8 +36,41 @@ func (help *Help) updateMap(){
     }
 }
 
+func (help *Help) autocorrectName(name string) string {
+    switch name {
+        // spells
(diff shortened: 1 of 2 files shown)


lib/lbx/help.go
```diff
@@ -36,12 +36,45 @@ func (help *Help) updateMap(){
     }
 }
 
+func (help *Help) autocorrectName(name string) string {
```
kazzmir, line 39: ok lets do this for now. we can also possibly do a levenshtein distance (aka edit distance) to find the closest match, and if the distance <2 or something then just choose that help entry

kazzmir:
btw at some point I think help.go should go into game/magic/help. I'm not really sure why I put it into lib/lbx, but lets do that in a different pr

kazzmir (approved):

PR #260, opened 2025-01-28
spellbook: simplify sorting

msom:
... I didn't realize sorting by name wasn't even necessary as `MagicType` is an `int` anyway.


kazzmir:
oh yea true, btw the go docs kind of suggest to use the newer slices.Sort functions. I use them where I can. Its not a big deal in any case https://pkg.go.dev/sort#Slice

kazzmir (approved):

msom:
good to know 👍

28.01.2025 05:58:55 Jon Rafkind ***@***.***>:

PR #263, opened 2025-01-29
map: refactor nodes and encounters

msom:
Use `ExtraEncounter` for all kinds of enemy encounters including the (initial) enemies on nodes.

Once an encounter is won, it will be removed from the map. Nodes initially have both an `ExtraEncounter` and an `ExtraMagicNode`, but the first player to win removes the encounter.

This will remove some code code duplication and makes some parts easier (for example path finding will never cross an encounter; removes `IsEmpty` as it isn't needed anymore).

It will also allow to add a set of players already explored an encounter centrally to `ExtraEncounter`.

### DIFF
--- a/game/magic/maplib/map.go
+++ b/game/magic/maplib/map.go
@@ -185,7 +185,7 @@ const (
 
 func (encounterType EncounterType) Name() string {
     switch encounterType {
-        case EncounterTypeLair: return "Monstar Lair"
+        case EncounterTypeLair: return "Monster Lair"
         case EncounterTypeCave: return "Mysterious Cave"
         case EncounterTypePlaneTower: return "Tower"
         case EncounterTypePlaneTowerOpen: return "Tower"
--- a/game/magic/maplib/map.go
+++ b/game/magic/maplib/map.go
@@ -178,13 +178,48 @@ const (
     EncounterTypeRuins
     EncounterTypeAbandonedKeep
     EncounterTypeDungeon
-
-    // somewhat of a hack, but its useful to have a single type that can represent all the
-    // different types of nodes and encounters
     EncounterTypeChaosNode
     EncounterTypeNatureNode
     EncounterTypeSorceryNode
 )
+
+func (encounterType EncounterType) Name() string {
+    switch encounterType {
+        case EncounterTypeLair: return "Monstar Lair"
+        case EncounterTypeCave: return "Mysterious Cave"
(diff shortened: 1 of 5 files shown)


game/magic/maplib/map.go
```diff
@@ -178,14 +178,49 @@ const (
...
-    // somewhat of a hack, but its useful to have a single type that can represent all the
-    // different types of nodes and encounters
     EncounterTypeChaosNode
     EncounterTypeNatureNode
     EncounterTypeSorceryNode
 )
 
+func (encounterType EncounterType) Name() string {
+    switch encounterType {
+        case EncounterTypeLair: return "Monstar Lair"
```
kazzmir, line 188: monster?

msom: I've taken that from `confirmLairEncounter` (https://github.com/kazzmir/master-of-magic/pull/263/files#diff-bdd32801d87fed666fc98e821ed2fff37b0c3e879d040c66701320ae8353e5adL3486)

msom: Might be right? https://masterofmagic.fandom.com/wiki/Monster_Lair

kazzmir: I just meant the 'a' should be an 'e'

kazzmir:
Probably not the fault of this pr, but sometimes the magic nodes have nothing in them, which I think should never happen.

image

game/magic/maplib/map.go
msom: Ah yes!

msom:
I think `enemyChoice == zero` in `chooseGuardianAndSecondary` shouldn't be an option?

kazzmir:
I think MakeMagicNode should do something like
```
budget := ...
ok := false
for !ok {
  guardians, secondary = ...
  if len(guardians) == 0 {
     budget += 50
  } else {
    ok = true
  }
}
```
So that it keeps increasing the budget until it selects something.

kazzmir:
image

this one is ok, but I think in this case its supposed to just say 'You have found some Ruins." and not the part about 'Scouts have spotted...". And if there are no monsters then the dialog box should just have one button for 'Ok'

msom:
The original does this:
Bildschirmfoto 2025-01-29 um 07 34 13

And next comes the trease popup.

kazzmir:
ok cool, i can add that

kazzmir (approved):

PR #270, opened 2025-01-31
combat: improve fleeing

msom:
Improves fleeing from combat.

I thought this would only be a small change but it turned out to be much bigger 😬

Changes:
- Units have a chance to die while fleeing
  - normal units have a chance of 50%, heroes of 25% ([wiki](https://masterofmagic.fandom.com/wiki/Encounter_Zone))
  - units dying while fleeing during combat count as killed units and will lead to experience (not 100% sure this makes sense)
- Fled units are relocated on the map
  - Movement costs/remaining movement is ignored (I think this is how it's implemented in the original game, but not 100% sure)
  - Attackers are reset to their last positions
  - Defenders try to find an empty location in a radius of one tile which has no encounters/enemy cities/enemy stacks (I think this more or less how it was in the original if I remember correctly)
    - if no position is found, they die
    - if only water is found, the units not being able to swim or fly die
    - units dying this way do not count as killed and will not lead to experience (not 100% sure this makes sense) 
- Combat end screen now shows a retreat message and how many units are killed (if any)

Bildschirmfoto 2025-01-31 um 09 26 48

Not yet implemented/open
- immobilized units should always be killed when fleeing, may be best to wait until the web spell etc. has been implemented and do some sort of unit.IsImmobilized?
- fleeing is not used in strategic combat, not sure if this would make sense
- fleeing is not used yet in AI (didn't want to interfere your work) (I think enemies from encounters should never flee)
- difficulty settings is not considered when calculating the chance of dying as it is unclear how exactly it changes
- only the number of died units are reported, not which ones; not sure if it is really necessary to report every single unit
- in the original there is also a confirmation button when clicking on "flee" during combat; not sure if this is really necessary

### DIFF
--- a/game/magic/game/game.go
+++ b/game/magic/game/game.go
@@ -3011,16 +3011,8 @@ func (game *Game) doMoveFleeingDefender(player *playerlib.Player, stack *playerl
         }
     }
 
-    // kill whole stack if no position found
-    if len(positions) == 0 && len(waterPositions) == 0 {
-        for _, unit := range stack.Units() {
-            player.RemoveUnit(unit)
-        }
-        return
-    }
-
-    // kill units that can not move to water
-    if len(positions) == 0 {
+    // kill units that can not move to water if only water is available
+    if len(positions) == 0 && len(waterPositions) != 0 {
         for _, unit := range stack.Units() {
             if !unit.IsFlying() && !unit.IsSwimmer() {
                 player.RemoveUnit(unit)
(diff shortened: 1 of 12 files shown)


kazzmir:
nice. I wouldn't worry about the AI fleeing yet, we can do that later

game/magic/game/game.go
```diff
@@ -2956,6 +2956,91 @@ func (game *Game) doMoveCamera(yield coroutine.YieldFunc, x int, y int) {
...
+        }
+        return
+    }
+
+    // set to a random position
+    position := positions[rand.IntN(len(positions))]
+    stack.SetX(position.X)
+    stack.SetY(position.Y)
+
+    allStacks := player.FindAllStacks(position.X, position.Y, stack.Plane())
```
kazzmir, line 3038: I added this part because if a defender ends up on a tile that already contains a stack of units of the same player, then the newly moved defenders should become part of that stack

kazzmir (approved):

game/magic/game/game.go
msom: makes sense!

PR #271, opened 2025-01-31
game: add some sources for fame

msom:
Adds some [sources for fame](https://masterofmagic.fandom.com/wiki/Fame#Sources_for_Fame):
- from Famous ability
- from battle
  - ±1 for 4+ enemies
  - ±1 for fantastic creatures with spell cost ≥ 600 MP. I moved the costs from `nodes` to `units` and added the missing costs (based on the [wiki](https://masterofmagic.fandom.com/wiki/Fantastic_Unit#List_of_Summoned_Units)). Theoretically, these costs could differ in the `spelldat.lbx` but I don't think it make much sense to adjust these when loading the lbx, right?
  - -1 for every two hero levels of a dead hero (starting from Myrmidon)

The following are still missing:
- from razing cities (added fixmes)
- from just cause spell (added a fixme)
- from defeated enemy wizard (that's not implemented yet, right?)

The gained / lost fame is now displayed in the combat end screen:
Bildschirmfoto 2025-01-31 um 21 13 08

### DIFF
--- a/game/magic/maplib/node.go
+++ b/game/magic/maplib/node.go
@@ -192,17 +192,17 @@ func computeNatureNodeEnemies(budget int) ([]units.Unit, []units.Unit) {
 
     enemyCosts := map[Enemy]int{
         None: 0,
-        WarBear: 70,
-        Sprite: 100,
-        EarthElemental: 160,
-        Spiders: 200,
-        Cockatrice: 275,
-        Basilisk: 325,
-        StoneGiant: 450,
-        Gorgons: 600,
-        Behemoth: 700,
-        Colossus: 800,
-        GreatWyrm: 1000,
+        WarBear: units.WarBear.CastingCost,
+        Sprite: units.Sprite.CastingCost,
+        EarthElemental: units.EarthElemental.CastingCost,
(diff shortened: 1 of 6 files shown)


msom:
I had to move the combat end screen part further below as it now requires the fame calculation (but this is also required for tactical combat) which is a bit ugly...

game/magic/maplib/node.go
```diff
@@ -192,17 +192,17 @@ func computeNatureNodeEnemies(budget int) ([]units.Unit, []units.Unit) {
...
-        EarthElemental: 160,
-        Spiders: 200,
-        Cockatrice: 275,
-        Basilisk: 325,
-        StoneGiant: 450,
-        Gorgons: 600,
-        Behemoth: 700,
-        Colossus: 800,
-        GreatWyrm: 1000,
+        WarBear: units.WarBear.CastingCost,
```
kazzmir, line 195: do these casting costs also line up with the spell cost for casting the summoning spells?

msom: I didn't check all but the ones I checked do...

kazzmir (approved):

PR #273, opened 2025-02-01
combat: remove killed defenders from encounters

msom:


game/magic/game/game.go
```diff
@@ -3730,8 +3730,13 @@ func (game *Game) doEncounter(yield coroutine.YieldFunc, player *playerlib.Playe
 
         game.createTreasure(encounter.Type, encounter.Budget, player)
     } else {
-        encounter.ExploredBy.Insert(player)
-        // FIXME: remove killed defenders
+        var remaining []units.Unit
+        for index := range enemies {
+            if enemies[index].GetHealth() != 0 {
```
kazzmir, line 3735: just for safety can this be `>` instead of `!=`

kazzmir: also just to note that if the last defender in the encounter fights with the last attacker and both units die in the melee combat, the defender wins the battle because its a tie. but then the encounter would be left with 0 units.

kazzmir: which I think is fine as far as this combat stuff goes, it just means a new attacker would find an empty encounter and immediately defeat it due to the code at the top of this method. but the surveyor probably needs to handle the case where an encounter has 0 units but has been explored

kazzmir: maybe the surveyor can say `Empty` for an explored but empty encounter?

kazzmir (approved):

kazzmir:
I guess an interesting consequence of this is that there could be an encounter with very strong units (say 2 sky drakes in it, or 2 great wyrms), where player1 attacks the encounter and defeats all but one of the minor guardians. Then player2 comes along and attacks, and only has to face the one minor guardian, so they win the encounter easily. The treasure player2 would get would be quite large, since the budget of the original encounter would be fairly high.

I guess this is how the game works!

PR #274, opened 2025-02-01
game: implement range of sight

msom:
Implements range of sight of units/cities and fog of war for units (but not terrain, cities etc.)

The fog map of a player can now have 3 values:
- unexplored: player has never seen the tile
- explored: player has once seen the tile
- visible: the tile is currently visible by a unit or city of a player

The wiki is very unclear about this topic. What I've found out from testing with the original and the wiki:
- Fog of war (explored tiles) hides enemy units
- Cities have a range of sight
  - normal cities have 2 square
  - cities with oracle have 4 square
  - cities with Nature's Eye spell have 5 square
- Units have a range of sight
  - normal units have 1 square
  - flying units have 2 square
  - units with scouting have 2 or 3 square (there seems to be no Scouting I, only II and III see [wiki](https://masterofmagic.fandom.com/wiki/Scouting), I changed the unit definitions accordingly)
- The start location gets an initial lift of 3 tile circle
- Nature's Awareness spell would make every tile visible

What I didn't change
- find path still considers explored tiles, not visible (not sure what the right strategy is here)
- surveyor still considers explored tiles, not visible (I think this is correct, maybe even for cities)
- spell location selection still consider explored tiles, not visible (I think this is correct)

I didn't implement anything regarding cities and terrain. This would be such a pain and probably also overkill and I'm not sure what the original does. To be honest, I would leave it like this and only do fog of war for units.

Fog of war is currently rendered with a transparent overlay both in the overworld and the minimap. Maybe add a setting for it?
Bildschirmfoto 2025-02-01 um 16 05 07

### DIFF
--- a/game/magic/game/game.go
+++ b/game/magic/game/game.go
@@ -6035,6 +6035,15 @@ func (overworld *Overworld) DrawFog(screen *ebiten.Image, geom ebiten.GeoM){
 
     // log.Printf("fog min %v, %v max %v, %v", minX, minY, maxX, maxY)
 
+    black := ebiten.ColorScale{}
+    black.Scale(1, 1, 1, 1)
+
+    lightTransparent := ebiten.ColorScale{}
+    lightTransparent.Scale(1, 1, 1, 0.3)
+
+    darkTransparent := ebiten.ColorScale{}
+    darkTransparent.Scale(1, 1, 1, 0.5)
+
     for x := minX; x < maxX; x++ {
         for y := minY; y < maxY; y++ {
             tileX := overworld.Map.WrapX(x)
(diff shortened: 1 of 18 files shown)


game/magic/game/game.go
```diff
@@ -1846,7 +1836,8 @@ func (game *Game) FindPath(oldX int, oldY int, newX int, newY int, stack *player
         }
 
         // don't know what the cost is, assume we can move there
-        if x2 >= 0 && x2 < len(fog) && y2 >= 0 && y2 < len(fog[x2]) && !fog[x2][y2] {
+        // FIXME: how should this behave for different FogTypes?
+        if x2 >= 0 && x2 < len(fog) && y2 >= 0 && y2 < len(fog[x2]) && fog[x2][y2] != data.FogTypeUnexplored {
```
kazzmir, line 1840: shouldn't this have been `==` instead of `!=` since the question condition is 'if the unit does not know about this tile then we don't know the cost'

msom: you're right

game/magic/maplib/map.go
```diff
@@ -1554,6 +1555,11 @@ func (mapObject *Map) DrawMinimap(screen *ebiten.Image, cities []MiniMapCity, ce
                 use = cityColor
             }
 
+            // FIXME: make this drawing fog of war configurable?
+            if fog[tileX][tileY] == data.FogTypeExplored {
+                use =  util.ToRGBA(util.Lighten(use, -50))
```
kazzmir, line 1560: we have to precompute these colors in a table, because Lighten() does some matrix math. With this if statement the cpu usage goes up ~20%. The cpu profile file shows this as the top usage of cpu time
```
1290ms 13.95% 13.95%     1290ms 13.95%  github.com/hajimehoshi/ebiten/v2/internal/affine.mulSquare (inline)
```

kazzmir:
very good overall!

game/magic/maplib/map.go
kazzmir: to make this simple we can use a caching function that takes two arguments (TerrainType and FogType) and returns a color. Then this code would be
```
use := getMapColor(terrain, fog)
```

kazzmir: I pushed code to do this, sorry if I we may have overlapped

msom: no worries, thanks!

kazzmir (approved):

PR #277, opened 2025-02-02
enchantment: show help

msom:
Shows help on right click for global and city enchantments. Like in the original game, only shows the help about global enchantments if no global enchantment are in place.


kazzmir:
I fixed a small bug when the last enchantment is canceled in the magic ui, then right clicking on the enchantment rect should bring up the generic enchantments help

kazzmir (approved):

PR #278, opened 2025-02-02
cast: implement wall of darkness

msom:
Implements "Wall of Darkness" Spell.

I also added most of the city enchantments definitions (names, upkeep cost, icon lbx index, sound index). Some values are still missing but I guess it's easiest to fix them when implementing the individual spells.

Also added the city enchantment upkeep costs to the total enchantment upkeep and dissipation of city enchantments.


game/magic/player/player.go
```diff
@@ -501,7 +501,13 @@ func (player *Player) TotalEnchantmentUpkeep() int {
         upkeep += enchantment.UpkeepMana()
     }
 
-    // FIXME: add city enchantments and unit enchantments
+    for _, city := range player.Cities {
```
kazzmir, line 504: This should include enchantments cast in other players' cities too right? Maybe this function could accept an interface that has a method for getting all city enchantments cast by a specific player, and then the Game type can implement that method

game/magic/player/player.go
```diff
@@ -501,7 +501,13 @@ func (player *Player) TotalEnchantmentUpkeep() int {
...
+        for _, enchantment := range city.Enchantments.Values() {
```
kazzmir, line 505: Also this needs to check the owner of the enchantment

msom: It does (`enchantment.Owner == player.GetBanner()`)

kazzmir: I only see the banner check in the dissipate enchantment code in game.go, but not here in the total upkeep code in player.go

kazzmir: We could have city.GetEnchantmentsCastBy(player) to just return the ones owned by that one player

msom: Ah yes, you're right

game/magic/player/player.go
msom: I added `players []*Player` but I can also change this to passing an interace

game/magic/game/game.go
```diff
@@ -5660,7 +5660,35 @@ func (game *Game) DissipateEnchantments(player *playerlib.Player, power int) {
         player.GlobalEnchantments.Remove(enchantment)
     }
 
-    // FIXME: dissipate unit enchantments and city enchantments
+    // keep removing city enchantments until there is no more mana issue
+    type CityEnchantment struct {
+        City citylib.City
```
kazzmir, line 5665: This should probably be *citylib.City

game/magic/player/player.go
```diff
@@ -494,23 +494,31 @@ func (player *Player) FoodPerTurn() int {
     return food
 }
 
-func (player *Player) TotalEnchantmentUpkeep() int {
+func (player *Player) TotalEnchantmentUpkeep(players []*Player) int {
```
kazzmir, line 497: An interface might be cleaner, because I think you can use the same method in the dissipate enchantment code when it gets all the city enchantments. The return type of the method can be a slice of a type that includes the city and an enchantment

msom: Right, this is much cleaner

kazzmir:
Looks good, I'll merge it soon

kazzmir (approved):

PR #282, opened 2025-02-03
util: use autocache in map- and terrainviewer

msom:


kazzmir:
lol I was just thinking that today.. util/mapeditor could use this too

msom:
it actually is the mapeditor 😂, my commit and pr title is wrong... i'll also add it to the terrainviewer

kazzmir (approved):

PR #283, opened 2025-02-03
Cast prosperity

msom:
I'm not 100% sure about the bonus calculation, I went with ["Gold is increased by 100%, like a Merchants' Guild does"](https://masterofmagic.fandom.com/wiki/Prosperity).

### DIFF
--- a/game/magic/city/city.go
+++ b/game/magic/city/city.go
@@ -954,6 +954,14 @@ func (city *City) GoldMerchantsGuild() int {
     return 0
 }
 
+func (city *City) GoldProsperity() int {
+    if city.HasEnchantment(data.CityEnchantmentProsperity) {
+        return city.GoldTaxation() + city.GoldMinerals()
+    }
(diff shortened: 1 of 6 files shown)


kazzmir:
image

do you know if this view in the city screen is updated? If you left click on the gold icons in the city view it brings up this ui, where it breaks down all the sources of gold. Should there be an additional line that shows the prosperity amount?

kazzmir:
The code for that screen is in cityview/city-screen.go:GoldProducers, around line 1428

kazzmir:
actually the comments on this wiki page shows that prosperity does show up as a line item
https://masterofmagic.fandom.com/wiki/Prosperity

msom:
Would be nice though if it showed up

msom:
Bildschirmfoto 2025-02-03 um 06 18 17

kazzmir (approved):

PR #289, opened 2025-02-04
cityview: fix myrror city display

msom:
Adds the sprite indexes for myrror cities:
Bildschirmfoto 2025-02-04 um 18 44 30

Also renders enchantments changing the background of the city (for both planes; in the order of precedence):
- flying fortress
- famine
- cursed lands
- gaias blessing

Or the horizon (for both planes; in the order of precedence):
- Chaos rift
- Heavily Light
- Cloud of Shadow

I also added some fixme's (where I'm quite sure but not 100%):
- hills and moutain horizon seem to depend on the tiles north of the city
- besides river there is also shore (I think an any sourrounding tile)

`test/cityview` allows to now to toggle plane (`p`-key) and the enchantments mentioned above (keys `1`-`7`).

### DIFF
--- a/test/cityview/main.go
+++ b/test/cityview/main.go
@@ -126,14 +126,37 @@ func NewEngine() (*Engine, error) {
     }, nil
 }
 
+func (engine *Engine) togglePlane() {
+    if engine.CityScreen.City.Plane == data.PlaneArcanus {
+        engine.CityScreen.City.Plane = data.PlaneMyrror
+    } else {
+        engine.CityScreen.City.Plane = data.PlaneArcanus
+    }
+}
+
+func (engine *Engine) toggleEnchantment(enchantment data.CityEnchantment) {
+    if engine.CityScreen.City.HasEnchantment(enchantment) {
+        engine.CityScreen.City.RemoveEnchantment(enchantment, data.BannerBlue)
+    } else {
+        engine.CityScreen.City.AddEnchantment(enchantment, data.BannerBlue)
+    }
+}
+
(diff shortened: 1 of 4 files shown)


kazzmir:
As I press the number keys should the enchantments window update? Right now these are static

image

msom:
you're right, these don't update, only the city scape

msom:
what is missing? i just Add/RemoveEnchantment

test/cityview/main.go
```diff
@@ -126,14 +126,38 @@ func NewEngine() (*Engine, error) {
...
+    } else {
+        engine.CityScreen.City.Plane = data.PlaneArcanus
+    }
+}
+
+func (engine *Engine) toggleEnchantment(enchantment data.CityEnchantment) {
+    if engine.CityScreen.City.HasEnchantment(enchantment) {
+        engine.CityScreen.City.RemoveEnchantment(enchantment, data.BannerBlue)
+    } else {
+        engine.CityScreen.City.AddEnchantment(enchantment, data.BannerBlue)
```
kazzmir, line 141: I think after an enchantment is added/removed that this code should basically do `engine.CityScreen.UI = engine.CityScreen.MakeUI(...)`

kazzmir: it looks like theres a few times that the UI field of the city screen is updated in city-screen.go, so maybe the CityScreen object can just have an UpdateUI() method that does the right thing. Probably it should always pass in buildinglib.BuildingNone to MakeUI

kazzmir (approved):

PR #293, opened 2025-02-05
cast: add cursed lands

msom:
Adds the spell and enchantment "cursed lands".

It has the following effect, [see](https://masterofmagic.fandom.com/wiki/Cursed_Lands):
- the absolute unrest is increase by 1 (following [this calculation](https://masterofmagic.fandom.com/wiki/Unrest#Sources_of_Unrest))
- the production rate is halved

Also fixes:
- `ProductionInspirations` not checking for the enchantment
- Worker production in the city view

I also started enabling (own) town curses in the simplified city view but got stuck with these issues:
- How to refresh after enchantment has been removed?
- How to avoid left clicking on an enchantment closing the view?


game/magic/cityview/city-screen.go
```diff
@@ -2016,12 +2017,34 @@ func SimplifiedView(cache *lbx.LbxCache, city *citylib.City, player *playerlib.P
...
             Draw: func(element *uilib.UIElement, screen *ebiten.Image){
                 var scale ebiten.ColorScale
                 scale.ScaleAlpha(getAlpha())
                 useFont.Print(screen, x, y, float64(data.ScreenScale), scale, enchantment.Enchantment.Name())
             },
+            LeftClick: func(element *uilib.UIElement) {
+                if enchantment.Owner == player.GetBanner() {
+                    yes := func(){
+                        city.RemoveEnchantment(enchantment.Enchantment, enchantment.Owner)
+                        // FIXME: how to refresh similar to  cityScreen.UI = cityScreen.MakeUI(buildinglib.BuildingNone)?
```
kazzmir, line 2032: I mean you can literally just do that here. it might be cleaner to add a CityScreen.RefreshUI() that updates the UI field, but you can just invoke RefreshUI() here with no issues I think

kazzmir: oh sorry is it because there is no reference to a cityScreen variable here?

msom: yes

kazzmir: Ok I get it, well I guess we would need some huge `makeUi := func(){ ... }` with all this code in it, and that function removes all ui elements from the `ui` object and re-adds all the new ones

kazzmir:
> How to avoid left clicking on an enchantment closing the view?

By adding a `LeftClick` field to the UI object itself, if a left click occurs but no element receives the left click then the ui's LeftClick func will fire. In this case I made the ui's LeftClick just set a quit=true variable

> How to refresh after enchantment has been removed?

I implemented the `setupUI()` trick that recreates all the ui elements after a city enchantment has been removed.

kazzmir (approved):
This PR looks good to me, anything else you want to change before merging?

msom:
Nice, thanks a lot!

The rest should be good!

PR #299, opened 2025-02-07
city: various fixes

msom:
I added a test for enchantment combinations and started testing city calculations against cities in Dos MoM and noticed some differences.

The following have been fixed:
- Granary has [+2 influence](https://masterofmagic.fandom.com/wiki/Granary) on maximum population instead of +1
- Production from farmers should be [rounded up](https://masterofmagic.fandom.com/wiki/Production#Base_Production)
- Add influcence of buildn [housing](https://masterofmagic.fandom.com/wiki/Housing) on population growth
- The influence of gaia's blessing is reported differently across the wiki, I changed so that it get's accounted for in [`BaseFoodLevel`](https://masterofmagic.fandom.com/wiki/Food#Base_Food_Level) and [excluded Wild Game](https://masterofmagic.fandom.com/wiki/Maximum_Population#Wild_Game)

The following fixmes have been added:
- The [fortress](https://masterofmagic.fandom.com/wiki/Fortress) should be "producing" the power from the spell books, this is currently in
-missing influences on population growth

Now, all the tested values either match or are covered by a FIXME in the code. The only exception is the growth rate: Here something seems wrong in the wiki, either the calculation of the maximum city size or the calculation of the growth rate (both seem to be lower in Dos MoM).

### DIFF
--- a/game/magic/city/city.go
+++ b/game/magic/city/city.go
@@ -943,6 +943,22 @@ func (city *City) PopulationGrowthRate() int {
 
     // FIXME: Add Famine, Stream of Life, Dark Rituals and Population Boom event
 
+    if city.ProducingBuilding == buildinglib.BuildingHousing {
+        bonus := 50
+        if city.Population > 1 {
+            bonus = (city.Workers / city.Population) * 100
+        }
+
+        if city.Buildings.Contains(buildinglib.BuildingBuildersHall) {
+            bonus += 15
+        }
+
+        if city.Buildings.Contains(buildinglib.BuildingSawmill) {
+            bonus += 10
+        }
+
+        base += bonus
+    }
 
     if city.SurplusFood() < 0 {
         base = 50 * city.SurplusFood()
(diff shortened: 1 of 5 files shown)


game/magic/city/city.go
```diff
@@ -595,6 +595,7 @@ func (city *City) ComputePower() int {
             case buildinglib.BuildingAlchemistsGuild: power += 3
             case buildinglib.BuildingWizardsGuild: power -= 3
             case buildinglib.BuildingFortress:
+                // FIXME: Wizards spell books should be reported in the city with the fortress
```
kazzmir, line 598: Any reason to leave this as a fixme? Can the total books be passed into city.compute power() as an argument?

kazzmir:
Btw eventually I think the 1/2 tile penalty will be an attribute of FullTile. We can deal with that later

game/magic/city/city.go
msom: Sure!

kazzmir (approved):

PR #301, opened 2025-02-08
cast: add pestilence

msom:
Adds [pestilence](https://masterofmagic.fandom.com/wiki/Pestilence) spell and enchantment.

Pestilence:
1. Adds +2 unrest
2. Removes 1000 population if a city contains 11 or more citizen
3. Randomly removes 1000 population if a city contains 10 or less citizen with a chance of `citizens > rand.IntN(10) + 1`

Pestilence seems to have no visual indication in the city screen and the original dos mom only shows the city screen when casting, so I done that.

(The wiki talks of 9 or less citizen for the third effect but that would create a stop a 10 citizens which the the original game clearly does not have.)

### DIFF
--- a/game/magic/game/game.go
+++ b/game/magic/game/game.go
@@ -6004,19 +6004,6 @@ func (game *Game) doCityEnchantments(player *playerlib.Player) {
             }
         }
 
-        // lose city population due to pestilence
-        if city.HasEnchantment(data.CityEnchantmentPestilence) {
-            if city.Citizens() >= 11 || city.Citizens() > (rand.IntN(10) + 1) {
-                city.Population -= 1000
-                var garrison []units.StackUnit
-                stack := player.FindStack(city.X, city.Y, city.Plane)
-                if stack != nil {
-                    garrison = stack.Units()
-                }
-                city.ResetCitizens(garrison)
-            }
-        }
-
         // FIXME: Add chaos rift etc.
     }
 }
--- a/game/magic/game/game.go
+++ b/game/magic/game/game.go
@@ -5965,49 +5965,6 @@ func (game *Game) DissipateEnchantments(player *playerlib.Player, power int) {
     // FIXME: dissipate unit enchantments
 }
 
-// apply the effects of city enchantments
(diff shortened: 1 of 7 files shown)


game/magic/game/game.go
```diff
@@ -6002,6 +6003,21 @@ func (game *Game) doGaiasBlessing(player *playerlib.Player) {
                 }
             }
         }
+
+        // lose city population due to pestilence
+        if city.HasEnchantment(data.CityEnchantmentPestilence) {
+            if city.Citizens() >= 11 || city.Citizens() > (rand.IntN(10) + 1) {
+                city.Population -= 1000
```
msom, line 6010: I'm not quite sure about this part removing 1000 population, does this make sense?

game/magic/game/game.go
```diff
@@ -6002,6 +6003,21 @@ func (game *Game) doGaiasBlessing(player *playerlib.Player) {
...
+                var garrison []units.StackUnit
+                stack := player.FindStack(city.X, city.Y, city.Plane)
+                if stack != nil {
+                    garrison = stack.Units()
+                }
+                city.ResetCitizens(garrison)
```
msom, line 6016: I'm not quite sure about this part removing 1000 population, does this make sense?

kazzmir: can this code for pestilence be moved into city.DoNextTurn()? That function also handles the case where the population changes (city.go:1445). We also need to handle the case where the population drops to 0, in which case the town should be destroyed by emitting the CityEvengtOutpostDestroyedf{} event

msom: Yes, I can move that. The population will never drop to 0 (citizens > rand.IntN(10) + 1 will never be true for 1 citizen).

kazzmir: oh true

kazzmir: I suppose the gias blessing stuff can also go into the city if the current map is passed into city.DoNextTurn() along with the garrison

msom: Yes, that would be better

kazzmir (approved):

PR #302, opened 2025-02-09
city: remove city if shrunk to zero

msom:
I tested with Dos MoM (v 1.60) and it's actually possible to shrink a city to zero by building / buying a settler.

Dos MoM doesn't show any notification, but I think it might be better to show something similar to the deserted outpost, what dou you think?

Bildschirmfoto 2025-02-09 um 07 06 55

I also moved city capping and growth message generation below the unit production part because building settler will also influence the population.

### DIFF
--- a/game/magic/city/city.go
+++ b/game/magic/city/city.go
@@ -26,6 +26,9 @@ type CityEventPopulationGrowth struct {
     Grow bool
 }
 
+type CityEventCityAbandoned struct {
+}
+
 type CityEventNewUnit struct {
     Unit units.Unit
     WeaponBonus data.WeaponBonus
@@ -1435,6 +1438,10 @@ func (city *City) DoNextTurn(garrison []units.StackUnit, mapObject *maplib.Map) 
             }
         }
 
+        if math.Abs(float64(city.Population/1000 - oldPopulation/1000)) > 0 {
+            cityEvents = append(cityEvents, &CityEventPopulationGrowth{Size: (city.Population - oldPopulation)/1000, Grow: city.Population > oldPopulation})
+        }
+
         buildingCost := city.BuildingInfo.ProductionCost(city.ProducingBuilding)
 
         if buildingCost != 0 || !city.ProducingUnit.Equals(units.UnitNone) {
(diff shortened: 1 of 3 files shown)


game/magic/city/city.go
```diff
@@ -1465,6 +1457,15 @@ func (city *City) DoNextTurn(garrison []units.StackUnit, mapObject *maplib.Map)
                 }
             }
         }
+
+        if city.Population > city.MaximumCitySize() * 1000 {
+            city.Population = city.MaximumCitySize() * 1000
+        }
+
+        if math.Abs(float64(city.Population/1000 - oldPopulation/1000)) > 0 {
```
kazzmir, line 1465: won't this show a notification for population shrinkage when a settler unit is created? I thought in that case no notification message is shown

msom: right; i also just noticed that size is only the difference so I can't check that for removing the city

game/magic/game/game.go
```diff
@@ -6065,23 +6065,27 @@ func (game *Game) StartPlayerTurn(player *playerlib.Player) {
...
                     verb = "shrunk"
                 }
 
-                // FIXME: can citizens shrink to 0, in which case the city should be removed?
-
-                scrollEvent := GameEventScroll{
-                    Title: "CITY GROWTH",
-                    Text: fmt.Sprintf("%v has %v to a population of %v.", city.Name, verb, city.Citizens()),
-                }
+                if growthEvent.Size == 0 {
```
kazzmir, line 6075: The Size field is the delta from the old population to the new population, so this seems like it should just check if the city.Citizens() == 0 now

kazzmir:
btw the gaias blessing code is outside the `if city.Outpost` if block, which would imply that gaias blessing can have an effect on outposts, is that the case?

msom:
> btw the gaias blessing code is outside the `if city.Outpost` if block, which would imply that gaias blessing can have an effect on outposts, is that the case?

Seems like it: Dos Mom does convert desert to grassland when casting it on an outpost.

kazzmir (approved):

Issue #306, opened 2025-02-10
cityview: fine-grained layout


msom:
I think Dos MoM uses a more fine grained grid for layouting the buildings in the city scape, where one tile corresponds to a single house/tree. I think it looks something like this:

<img width="450" alt="Image" src="https://github.com/user-attachments/assets/d851b49a-d64b-457a-a8fa-a7bef92cb8e2" />

Dos MoM doesn't seem to bother how the old layout was when adding a new building, for example when adding a stable:

<img width="450" alt="Image" src="https://github.com/user-attachments/assets/087898bc-62fb-4b62-bd88-7165fa802620" />

<img width="450" alt="Image" src="https://github.com/user-attachments/assets/9bfce47d-7f89-472f-b4e7-1d662d90c18b" />

An exception might be when replacing a building (for example a shrine with a temple) and they have the same size. Also, Shipwright's Guild, Ship Yard and Maritime Guild are always placed at the middle left.

I think the sizes of the buildings are:

```
func (building Building) Size() (int, int) {
    switch building {
        case BuildingBarracks: return 2, 3
        case BuildingArmory: return 2, 2
        case BuildingFightersGuild: return 3, 2
        case BuildingArmorersGuild: return 4, 2
        case BuildingWarCollege: return 3, 2
        case BuildingSmithy: return 2, 2
        case BuildingStables: return 3, 3
        case BuildingFantasticStable: return 3, 3
        case BuildingAnimistsGuild: return 2, 2
        case BuildingSawmill: return 2, 2
        case BuildingLibrary: return 3, 2
        case BuildingUniversity: return 3, 2
        case BuildingSagesGuild: return 2, 2
        case BuildingOracle: return 2, 2
        case BuildingAlchemistsGuild: return 1, 1
        case BuildingWizardsGuild: return 2, 2
        case BuildingShrine: return 2, 2
        case BuildingTemple: return 3, 3
        case BuildingParthenon: return 3, 3
        case BuildingCathedral: return 3, 3
        case BuildingMarketplace: return 2, 2
        case BuildingBank: return 2, 2
        case BuildingMerchantsGuild: return 2, 2
        case BuildingGranary: return 2, 2
        case BuildingFarmersMarket: return 2, 2
        case BuildingForestersGuild: return 2, 2
        case BuildingBuildersHall: return 2, 2
        case BuildingMechaniciansGuild: return 2, 2
        case BuildingMinersGuild: return 2, 1
[...]

PR #307, opened 2025-02-10
cityview: don't render replaced buildings

msom:
Dos MoM doesn't render some of the buildings and shows them as "Replaced" in the resource dialog.

Replaced buildings

This will free up some city slots. I think at some point it might be required to switch to a more fine-grained grid layout similar to what Dos MoM does, as the building slots will still run out quickly for bigger cities. I've added some screen shots for reference #306.

### DIFF
--- a/game/magic/building/building.go
+++ b/game/magic/building/building.go
@@ -130,3 +130,25 @@ func GetBuildingIndex(building Building) int {
 
     return -1
 }
+
+// the building which is shown in the city scape instead
+func (building Building) ReplacedBy() Building {
+    switch building {
+        case BuildingBarracks: return BuildingArmory
+        case BuildingFightersGuild: return BuildingArmorersGuild
+        case BuildingArmorersGuild: return BuildingWarCollege
+        case BuildingStables: return BuildingFantasticStable
+        case BuildingLibrary: return BuildingUniversity
+        case BuildingAlchemistsGuild: return BuildingWizardsGuild
+        case BuildingShrine: return BuildingTemple
+        case BuildingTemple: return BuildingParthenon
+        case BuildingParthenon: return BuildingCathedral
+        case BuildingMarketplace: return BuildingBank
+        case BuildingBank: return BuildingMerchantsGuild
+        case BuildingGranary: return BuildingFarmersMarket
+        case BuildingShipwrightsGuild: return BuildingShipYard
+        case BuildingShipYard: return BuildingMaritimeGuild
+    }
+
+    return BuildingNone
+}
--- a/game/magic/building/building.go
+++ b/game/magic/building/building.go
@@ -164,7 +164,7 @@ func (building Building) Size() (int, int) {
         case BuildingSmithy: return 2, 2
         case BuildingStables: return 3, 3
         case BuildingFantasticStable: return 3, 3
-        case BuildingAnimistsGuild: return 0, 0
+        case BuildingAnimistsGuild: return 2, 2
         case BuildingSawmill: return 2, 2
         case BuildingLibrary: return 3, 2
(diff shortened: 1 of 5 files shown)


game/magic/building/building.go
```diff
@@ -130,3 +130,64 @@ func GetBuildingIndex(building Building) int {
...
+    switch building {
+        case BuildingBarracks: return 2, 3
+        case BuildingArmory: return 2, 2
+        case BuildingFightersGuild: return 3, 2
+        case BuildingArmorersGuild: return 4, 2
+        case BuildingWarCollege: return 3, 2
+        case BuildingSmithy: return 2, 2
+        case BuildingStables: return 3, 3
+        case BuildingFantasticStable: return 3, 3
+        case BuildingAnimistsGuild: return 0, 0
```
kazzmir, line 167: 0,0?

game/magic/cityview/city-screen.go
```diff
@@ -403,6 +413,15 @@ func sellAmount(city *citylib.City, building buildinglib.Building) int {
     return cost
 }
 
+func wasBuildingReplaced(building buildinglib.Building, city *citylib.City) bool {
+    if building == buildinglib.BuildingNone {
+        return false
+    }
+
+    replacedBy := building.ReplacedBy()
+    return replacedBy != buildinglib.BuildingNone && city.Buildings.Contains(replacedBy)
```
kazzmir, line 422: we can consider handling the case that an intermediate replacement building was removed. For example, a shrine is replaced by a temple and a temple is replaced by a parthenon. If the temple was destroyed, the shrine still shouldn't show up as long as the parthenon still exists.

I think dos mom will simply refuse to destroy buildings that have been replaced, but I think that might be a mistake and we should allow it.

kazzmir: I guess the transitive case can be `replacedBy != buildinglib.BuildingNone && (city.Buildings.Containse(replacedBy) || wasBuildingReplaced(replacedBy, city)`

game/magic/cityview/city-screen.go
```diff
@@ -1437,6 +1460,7 @@ func (cityScreen *CityScreen) MakeResourceDialog(title string, smallIcon *ebiten
         }
     }
 
+    // FIXME: There can be a second page and a "More" button on the bottom right
```
kazzmir, line 1463: do you have a picture of what that 'more' button looks like in dos mom?

msom:
Thank's for the feedback! I'll look into it later (in the CET-evening)

game/magic/building/building.go
msom: fixed

game/magic/cityview/city-screen.go
msom: Bildschirmfoto 2025-02-09 um 18 13 50

Bildschirmfoto 2025-02-09 um 18 14 04

Clicking on the second "More" will show the first again

game/magic/building/building.go
```diff
@@ -130,3 +130,64 @@ func GetBuildingIndex(building Building) int {
...
+        case BuildingShipYard: return BuildingMaritimeGuild
+    }
+
+    return BuildingNone
+}
+
+// the size of the picture for this building (in squares)
+func (building Building) Size() (int, int) {
+    switch building {
+        case BuildingBarracks: return 2, 3
```
kazzmir, line 159: How did you figure out what these numbers should be? Is it just an estimate for each building?

msom: Yes, it's only an estimate. It's more or less recognizable if you take the small houses (in the other big grids) as a reference...

kazzmir (approved):

PR #308, opened 2025-02-10
cityview: use finer grained layout for buildings in the cityscape

kazzmir:
This is an experiment to use a finer grained method to layout buildings in the cityscape for https://github.com/kazzmir/master-of-magic/issues/306

The cityscape land is split up into rectangles that represent a patch of land, and the algorithm uses a recursive backtracking algorithm (similar to N-queens) to place each building on some patch of land. If some building can't be placed in a rectangle then the algorithm backtracks to put the previous building in a different rectangle than the one it was placed in and the algorithm continues from there.

The total empty space of all patches of land is 158 and the total space of all used buildings is 150, but it turns out that not all buildings can be placed in the cityscape at once (at least, the algorithm wasn't able to find a solution after several minutes).

When the buildings are filtered such that buildings that were replaced are removed from the list then a solution can always be found without any backtracking in about ~40 microseconds on my computer.

The city could compute the positions of the buildings and cache it, until a new building is added or a building is destroyed, at which point new positions could be computed, but taking care to keep the positions of existing buildings the same.

The algorithm will likely have to be transformed to be loop-based instead of recursive as javascript/wasm has a very small stack and cannot tolerate many recursive calls.


kazzmir:
closes https://github.com/kazzmir/master-of-magic/issues/306

PR #313, opened 2025-02-12
cast: dark rituals

msom:
Adds the [Dark Rituals](https://masterofmagic.fandom.com/wiki/Dark_Rituals) spell.

Effects:
- doubles power from religious buildings
- increases unrest by 1
- reduces population growth by 25%

Changes in city screen:
- remove buildings / slots when disabling enchantments
- report power from dark rituals

Als oupdates / adds some fixmes for later.

### DIFF
--- a/game/magic/cityview/city-screen.go
+++ b/game/magic/cityview/city-screen.go
@@ -1041,12 +1041,9 @@ func (cityScreen *CityScreen) MakeUI(newBuilding buildinglib.Building) *uilib.UI
                         if ok {
                             cityScreen.City.Buildings.Remove(building)
 
-                            for index, slot := range cityScreen.Buildings {
-                                if slot.Building == building {
-                                    cityScreen.Buildings = append(cityScreen.Buildings[:index], cityScreen.Buildings[index+1:]...)
-                                    break
-                                }
-                            }
+                            cityScreen.Buildings = slices.DeleteFunc(cityScreen.Buildings, func(slot BuildingSlot) bool {
+                                return slot.Building == building
+                            })
                         }
 
                         cityScreen.UI = cityScreen.MakeUI(buildinglib.BuildingNone)
--- a/game/magic/cityview/city-screen.go
+++ b/game/magic/cityview/city-screen.go
@@ -2001,7 +2001,7 @@ func (cityScreen *CityScreen) PowerProducers() []ResourceUsage {
     var usage []ResourceUsage
 
     add := func(count int, name string, building buildinglib.Building){
-        if count != 0 && cityScreen.City.Buildings.Contains(building) {
+        if count != 0 && (building == buildinglib.BuildingNone || cityScreen.City.Buildings.Contains(building)) {
             usage = append(usage, ResourceUsage{
                 Count: count,
                 Name: name,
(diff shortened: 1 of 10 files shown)


game/magic/cityview/city-screen.go
```diff
@@ -1030,6 +1035,20 @@ func (cityScreen *CityScreen) MakeUI(newBuilding buildinglib.Building) *uilib.UI
                 if enchantment.Owner == cityScreen.Player.GetBanner() {
                     yes := func(){
                         cityScreen.City.RemoveEnchantment(enchantment.Enchantment, enchantment.Owner)
+
+                        enchantmentBuildings := enchantmentBuildings()
+                        building, ok := enchantmentBuildings[enchantment.Enchantment]
+                        if ok {
+                            cityScreen.City.Buildings.Remove(building)
+
+                            for index, slot := range cityScreen.Buildings {
```
kazzmir, line 1044: I think this can be simplified with slices.DeleteFunc

game/magic/cityview/city-screen.go
```diff
@@ -2001,6 +2020,10 @@ func (cityScreen *CityScreen) PowerProducers() []ResourceUsage {
     add(-3, "Wizard's Guild", buildinglib.BuildingWizardsGuild)
     add(cityScreen.City.PowerMinerals(), "Minerals", buildinglib.BuildingNone)
 
+    if cityScreen.City.HasEnchantment(data.CityEnchantmentDarkRituals) {
+        add(cityScreen.City.PowerDarkRituals(), "Dark Rituals", buildinglib.BuildingNone)
```
kazzmir, line 2024: this is ok but maybe it could be slightly more elegant with a function just for enchantments. `addEnchantment(..., "Dark Rituals")` so you aren't forced to pass in a useless BuildingNone argument

game/magic/cityview/city-screen.go
```diff
@@ -2430,6 +2453,8 @@ func SimplifiedView(cache *lbx.LbxCache, city *citylib.City, player *playerlib.P
 
                 localOptions.GeoM.Translate(float64(3 * data.ScreenScale), float64(-2 * data.ScreenScale))
 
+                // FIXME: Unrest is currently only updated at the start of the turn, not when selling
+                //        buildings or adding/removing enchantments
```
kazzmir, line 2457: This reminds me that in CityScreen.SellBuilding it also does not update the unrest for a city after selling a building. I think it is at least possible to do while selling a building because the current garrison is available.

kazzmir: Well I guess the player owning the city currently being viewed in this simplified view is also available, so we could update the unrest here as well when an enchantment is removed.

kazzmir: lets fix the unrest issues in a different pr

kazzmir:
I adjusted the view when a city enchantment is cast to fade in the enchantment building as well.

game/magic/game/game.go
```diff
@@ -3251,6 +3251,7 @@ func (game *Game) defeatCity(yield coroutine.YieldFunc, attacker *playerlib.Play
 
         city.Buildings.Remove(buildinglib.BuildingFortress)
         city.Buildings.Remove(buildinglib.BuildingSummoningCircle)
+        // FIXME: remove enchantments (and their buildings)?
```
kazzmir, line 3254: this is a good question. I would argue any enchantments cast on the city by the wizard that originally owned the city should be dispelled, but any other enchanments cast by other wizards should be kept.

PR #318, opened 2025-02-15
cast: fix floating island summoning

msom:
Floating island is actually not summoned to the summoning circle but cast on a map tile with water (including lakes).

Also includes some smaller fixes regarding summoning:
- only show combat spells in the spellbook during combat
- fixes summoning picture of some units: shadow demons, death knights, cockatrices, giant spiders

So far, only "Lycantrophy" is missing of all the summoning spells! Not sure how this is rendered though...

Maybe better add some helper function to look up a stack on map?

### DIFF
--- a/game/magic/game/cast.go
+++ b/game/magic/game/cast.go
@@ -31,6 +31,7 @@ type LocationType int
 const (
     LocationTypeAny LocationType = iota
     LocationTypeLand
+    LocationTypeEmptyWater
     LocationTypeFriendlyCity
     LocationTypeEnemyCity
     LocationTypeFriendlyUnit
@@ -59,7 +60,7 @@ func (game *Game) doCastSpell(player *playerlib.Player, spell spellbook.Spell) {
              "Unicorns", "Basilisk", "Behemoth", "Cockatrices", "Colossus",
              "Earth Elemental", "Giant Spiders", "Gorgons", "Great Wyrm",
              "Sprites", "Stone Giant", "War Bears", "Air Elemental",
-             "Djinn", "Floating Island", "Nagas", "Phantom Beast", "Phantom Warriors",
+             "Djinn", "Nagas", "Phantom Beast", "Phantom Warriors",
              "Sky Drake", "Storm Giant", "Chaos Spawn", "Chimeras", "Doom Bat",
              "Efreet", "Fire Elemental", "Fire Giant", "Gargoyles",
              "Great Drake", "Hell Hounds", "Hydra", "Death Knights",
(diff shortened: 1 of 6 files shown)


game/magic/game/cast.go
```diff
@@ -933,3 +959,15 @@ func (game *Game) doCastGlobalEnchantment(yield coroutine.YieldFunc, player *pla
     yield()
 
 }
+
+func (game *Game) doCastFloatingIsland(yield coroutine.YieldFunc, player *playerlib.Player, tileX int, tileY int) {
+    update := func (x int, y int, frame int) {
+        if frame == 5 {
+            overworldUnit := units.MakeOverworldUnitFromUnit(units.GetUnitByName("Floating Island"), tileX, tileY, game.CurrentMap().Plane, player.Wizard.Banner, player.MakeExperienceInfo())
```
kazzmir, line 966: I think its better to just use `units.FloatingIsland` instead of searching for it by name

kazzmir:
I think there are some other things to consider for floating island:
 * the floating island unit itself will never be part of combat
 * land-based units can be transported on the floating island over water, similar to the ships

game/magic/game/cast.go
```diff
@@ -479,6 +486,25 @@ func (game *Game) selectLocationForSpell(yield coroutine.YieldFunc, spell spellb
...
                             }
                         }
+                    case LocationTypeEmptyWater:
+                        if tileY >= 0 && tileY < overworld.Map.Map.Rows() {
+                            tileX = overworld.Map.WrapX(tileX)
+                            if player.IsTileExplored(tileX, tileY, game.Plane) {
+                                if overworld.Map.GetTile(tileX, tileY).Tile.IsWater() {
+                                    empty := true
+                                    for _, enemy := range game.GetEnemies(player) {
+                                        if enemy.FindStack(tileX, tileY, game.Plane) != nil {
```
kazzmir, line 496: this is ok for now but I think we will need a helper method for this because plane switching needs to check what stack is on the other side. I'll add that later

msom:
> I think there are some other things to consider for floating island:
> 
>     * the floating island unit itself will never be part of combat
> 
>     * land-based units can be transported on the floating island over water, similar to the ships

Oh, I didn't realize this wasn't implemented, I just thought only the summoning part was wrong...

game/magic/game/cast.go
msom: maybe it would even make sense to store the stacks for lookup? I saw they also get cached for path finding...

kazzmir:
> Oh, I didn't realize this wasn't implemented, I just thought only the summoning part was wrong...

This stuff can be a separate PR I guess. I think the transport stuff might be relatively straight forward. Theres a few places in game.go that check if the unit is sailing, but really the check should be if the unit has the AbilityTransport ability

game/magic/game/cast.go
kazzmir: oh yea I was thinking about being able to compute some sort of cache of all enemies/cities that can be used in some local context.
```
cache := game.ComputeStackCityLocations()
if cache.ContainsStack(x, y) ...
if cache.ContainsCity(x, y) ...
```
something like this. Also I was considering just storing the city values in a map[image.Point] instead of an array so that its not a linear lookup every time to find a city. Stacks on the other hand would be a bit painful to store in a map, as the map would constantly have to be updated.

kazzmir (approved):

PR #322, opened 2025-02-16
cast: add incarnation

msom:
Implements the "Incarnation" spell:

Bildschirmfoto 2025-02-16 um 07 43 44

Also fixes some small issues:
- `GameEventVault` missing player variable
- deceased heros keeping equipment (Torin can be incarnated again, other heros resurrected)
- level of hired heros getting reset when being hired (heros will level up when not being hired, torin keeps his level on re-summoning)

### DIFF
--- a/game/magic/game/cast.go
+++ b/game/magic/game/cast.go
@@ -255,9 +255,8 @@ func (game *Game) doCastSpell(player *playerlib.Player, spell spellbook.Spell) {
             game.doSummonHero(player, false)
         case "Summon Champion":
             game.doSummonHero(player, true)
-
-        // FIXME: Incarnation
-        // Incarnation	Icon LifeLife	Rare	500	--	12	960	Summons Torin the Chosen – one of the most powerful Champions in the game.
+        case "Incarnation":
+            game.doIncarnation(player)
         case "Enchant Road":
             game.Events <- &GameEventSelectLocationForSpell{Spell: spell, Player: player, LocationType: LocationTypeAny, SelectedFunc: game.doCastEnchantRoad}
         case "Corruption":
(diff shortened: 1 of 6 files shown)


game/magic/game/cast.go
```diff
@@ -388,6 +387,23 @@ func (game *Game) doSummonHero(player *playerlib.Player, champion bool) {
     }
 }
 
+func (game *Game) doIncarnation(player *playerlib.Player) {
+    for _, hero := range game.Heroes {
+        if hero.HeroType == herolib.HeroTorin && hero.Status != herolib.StatusEmployed {
+            event := GameEventHireHero{
```
kazzmir, line 393: btw the AI wizards can summon torin, and their torin is a different torin from the player's one: https://masterofmagic.fandom.com/wiki/Torin_the_Chosen#Resurrecting_Torin

I'm not totally sure how to handle it, so this is fine for now

game/magic/player/player.go
```diff
@@ -285,7 +285,12 @@ func (player *Player) AddHero(hero *herolib.Hero) bool {
         if player.Heroes[i] == nil {
             player.Heroes[i] = hero
 
-            hero.Unit = units.MakeOverworldUnitFromUnit(hero.GetRawUnit(), fortressCity.X, fortressCity.Y, fortressCity.Plane, player.Wizard.Banner, player.MakeExperienceInfo())
+            level := hero.GetExperienceLevel()
+            experienceInfo := player.MakeExperienceInfo()
+
+            hero.Unit = units.MakeOverworldUnitFromUnit(hero.GetRawUnit(), fortressCity.X, fortressCity.Y, fortressCity.Plane, player.Wizard.Banner, experienceInfo)
+            hero.AddExperience(level.ExperienceRequired(experienceInfo.HasWarlord(), experienceInfo.Crusade()))
```
kazzmir, line 292: can you add a comment here about why experience is added?

kazzmir:
I fixed a bug regarding the health of a hero, because torin usually has Constitution his max health is affected by abilities but the hero.AdjustHealth() was not considering the extra health from abilities

game/magic/game/cast.go
msom: I think the bigger issue is that every player has its own [set of heros](https://masterofmagic.fandom.com/wiki/Hero#Description), appearing under different names (for example [Theria/Shadow/Misty/Doria/Tamara](https://masterofmagic.fandom.com/wiki/Theria_the_Thief#Alternate_Names). I guess it should be quite easy to move the heros from `Game` to `Player` and adjust the code...?

kazzmir: oh ok, yea I think we can just move the heroes array to Player and it should be fine. That makes things simple for torin

kazzmir (approved):

game/magic/game/cast.go
msom: Seems like the hero names are in `NAMES.LBX`

PR #328, opened 2025-02-17
unitview: make ui elements for abilities/items to interact with

kazzmir:
Be able to click on items/abilities in the unit view

image


kazzmir:
This ended up being a large refactor in that a new ui.UIElementGroup object is defined that allows the code to add elements to the group, and the group can be removed all at once from the ui. This is much simpler than having to manually add/remove elements directly from a UI object.

PR #337, opened 2025-02-18
WIP Evil presence

msom:


kazzmir:
image

after destroying a building that has the evil presence overlay, the overlay still shows up. Should that be the case?

kazzmir:
Casting evil presence on a town with consecration:
image

probably the message needs to be adjusted. we can do that later

kazzmir (approved):

PR #339, opened 2025-02-18
Overworld: implement Consecration

sidav:
I don't really like game/magic/game/cast.go lines 113 - 120. Maybe there's a new func (something like RemoveAllCityEnchantmentsOfTypes()) needed? The current RemoveCityEnchantment requires providing the spell's owner, which is not the case with Consecration.
I also don't really like 6868-6880 in game.go. I think we need a func like useMap.performFuncForEachCityCatchmentTile() with city and func (tileX, tilyY int) as arguments?
Also, I've added chaos/death wards checks for the spells affected by Consecration-related changes, except for some cases where there already was one.

### DIFF
--- a/game/magic/game/game.go
+++ b/game/magic/game/game.go
@@ -6865,19 +6865,8 @@ func (game *Game) doCleanCorruptionForConsecratedCities() {
             continue
         }
         useMap := game.GetMap(city.Plane)
-        for dx := -2; dx <= 2; dx++ {
-            for dy := -2; dy <= 2; dy++ {
-                if int(math.Abs(float64(dx)) + math.Abs(float64(dy))) == 4 {
-                    continue
-                }
-                cx := useMap.WrapX(city.X + dx)
-                cy := city.Y + dy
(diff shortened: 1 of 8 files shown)


game/magic/game/game.go
```diff
@@ -6854,6 +6854,34 @@ func (game *Game) doEarthquake(city *citylib.City, player *playerlib.Player) (in
...
+    choices := game.AllCities()
+    if len(choices) == 0 {
+        return
+    }
+    for _, city := range choices {
+        if !city.HasEnchantment(data.CityEnchantmentConsecration) {
+            continue
+        }
+        useMap := game.GetMap(city.Plane)
+        for dx := -2; dx <= 2; dx++ {
```
kazzmir, line 6868: You could also do this as 
```
for point, _ := range useMap.GetCatchmentArea(city.X, city.Y) {
  useMap.RemoveCorruption(point.X, point.Y)
}
```

kazzmir (approved):

kazzmir:
:+1:

PR #341, opened 2025-02-18
fizzle somes overland spells

kazzmir:
Some overland spells fizzle if targeting a town with consecration: https://masterofmagic.fandom.com/wiki/Consecration


kazzmir:
actually lets just let https://github.com/kazzmir/master-of-magic/pull/339 implement this

PR #343, opened 2025-02-19
cast: add wall of stone

msom:


game/magic/game/game.go
```diff
@@ -4805,35 +4805,25 @@ func GetUnitImage(unit units.StackUnit, imageCache *util.ImageCache, banner data
...
+    var animationIndex int = 0
 
-    switch size {
-        case citylib.CitySizeHamlet: index = 0
-        case citylib.CitySizeVillage: index = 1
-        case citylib.CitySizeTown: index = 2
-        case citylib.CitySizeCity: index = 3
-        case citylib.CitySizeCapital: index = 4
+    if city.HasWall() {
+        spriteIndex = 20
```
kazzmir, line 4813: I've never actually seen a city with sprite index 21 in dos mom, which is why the NoWallImage function wasn't called. Have you seen a case where sprite 21 is used?

msom: Yes, I think it was a bug that has been (community) fixed:

Bildschirmfoto 2025-02-19 um 05 16 20

I quite like (especially since it's already in the assets)

kazzmir:
looks good to me. any other changes?

msom:
Yes, the wall of stone spell... but we can also merge this and I'll open a new PR

kazzmir:
ok, up to you. if you mark this as ready ill merge it, otherwise you can add the spell here

msom:
The spell is now also implemented. A small issue remains in that the city in the overworld changes to the walled image before the fade-in of the cast...

game/magic/game/cast.go
```diff
@@ -77,6 +79,19 @@ func (game *Game) doCastSpell(player *playerlib.Player, spell spellbook.Spell) {
...
             game.Events <- &GameEventSelectLocationForSpell{Spell: spell, Player: player, LocationType: LocationTypeEmptyWater, SelectedFunc: selected}
+        case "Wall of Stone":
+            selected := func (yield coroutine.YieldFunc, tileX int, tileY int){
+                chosenCity, _ := game.FindCity(tileX, tileY, game.Plane)
+                if chosenCity == nil {
+                    return
+                }
+
+                chosenCity.AddBuilding(building.BuildingCityWalls)
+                game.doCastCityEnchantment(yield, tileX, tileY, player, data.CityEnchantmentWallOfStone)
```
kazzmir, line 90: It seems like you wouldn't need to define the wall of stone enchantment if cityview.MakeEnchantmentView took in a building instead of an enchantment. Probably also that function would have to take in the name of the spell being cast. Then you could have a new function that calls MakeEnchantmentView but with a building instead. Most of the code in showCityEnchantment can be parameterized by the ui and quit variables

kazzmir:
Also there might be a small edge case where a city is building city walls but then the player casts wall of stone on the city. I'm not sure what dos mom does in that case but maybe the city could revert to trade goods?

game/magic/game/cast.go
msom: I had this implemented like this at one point but it felt like too much changes needed in the code compared to just adding a new enchantment....

kazzmir: To be more clear, there could be two methods
```
MakeEnchantmentViewForEnchantment(enchantment)
MakeEnchantmentViewForBuilding(building, name)
```
Where the ForBuilding one does basically what the MakeEnchantmentView() does now, and the ForEnchantment just invokes ForBuilding with the associated building and enchantment name

kazzmir: yea its kind of awkward either way, all this just to support this one unique enchantment. The check for BuildingCityWalls in layout seems strange though

kazzmir:
> The spell is now also implemented. A small issue remains in that the city in the overworld changes to the walled image before the fade-in of the city enchantment view appears, not really sure how to address that....

yea I don't see a way around this. I think its ok for now

game/magic/game/cast.go
msom: That's were i though it's actually clearer, because city walls doesn't need to be considered for layout since it has a fixed position...

game/magic/cityview/city-screen.go
```diff
@@ -297,7 +298,7 @@ func makeBuildingSlots(city *citylib.City) []BuildingSlot {
     enchantmentBuildings := enchantmentBuildings()
 
     for enchantment, building := range enchantmentBuildings {
-        if city.HasEnchantment(enchantment) {
+        if city.HasEnchantment(enchantment) && building != buildinglib.BuildingCityWalls {
```
kazzmir, line 301: actually you can make this more elegant by just removing city walls from toLayout unconditionally after this for loop. Basically line 289 can go after this for loop

kazzmir (approved):

PR #364, opened 2025-02-22
cast: add just cause

msom:
Adds [Just Cause](https://masterofmagic.fandom.com/wiki/Just_Cause):
- Adds +10 to basic fame 
- Adds +1 pacification

This PR also ensures that basic fame will never go below 10, so minimum fame will be 10 with Just Cause.

### DIFF
--- a/game/magic/city/city_test.go
+++ b/game/magic/city/city_test.go
@@ -11,6 +11,7 @@ import (
     "github.com/kazzmir/master-of-magic/game/magic/maplib"
     "github.com/kazzmir/master-of-magic/game/magic/units"
     "github.com/kazzmir/master-of-magic/lib/fraction"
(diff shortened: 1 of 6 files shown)


game/magic/city/city_test.go
```diff
@@ -97,6 +98,10 @@ func (reign *NoReign) GetBanner() data.BannerType {
     return data.BannerBlue
 }
 
+func (reign *NoReign) GetGlobalEnchantments() *set.Set[data.Enchantment] {
+    return nil
```
kazzmir, line 102: I think this can't be nil

kazzmir:
Just a note that the unrest in the towns doesn't get updated until the following turn after casting just cause because that is when UpdateUnrest(...) gets called in the NextTurn code. I think this is fine.

kazzmir (approved):

PR #365, opened 2025-02-22
city: update unrest on enchantment changes

msom:
Update unrest when adding and/or removing city and/or global enchantments.

I also added (the already existing) `GetUnits` to `ReignProvider` so cities can query their garrison themselves.

@kazzmir What do you think? Anything I missed?


kazzmir (approved):
ok

PR #367, opened 2025-02-22
Tranquility spell

sidav:


game/magic/game/game.go
```diff
@@ -5231,6 +5232,37 @@ func (game *Game) IsGlobalEnchantmentActive(enchantment data.Enchantment) bool {
     })
 }
 
+// Returns true if dispel is successful. This func is just a check and by itself doesn't change anything in the game.
+// TargetSpellCost may be either a base cost or an effective cost (https://masterofmagic.fandom.com/wiki/Casting_Cost#Dispelling_Magic)
+// FIXME: add Runemaster check here (it should increase the dispel strength)
+func (game *Game) RollDispelChance(dispelStrength int, targetSpellCost int, targetSpellRealm data.MagicType, targetSpellOwner *playerlib.Player, 
+    checkArchmageRetort bool, checkRealmMasteryRetorts bool) bool {
```
kazzmir, line 5239: is there a case where we would check archmage but not realm mastery or vice-versa?

kazzmir: maybe this function could just return the chance by itself so that we can write tests that compute the chance for various different situations, and other another function can do the rand.N(250) < chance

sidav: It seems you're right, realm mastery is always checked alongside archmage. Runemaster is not the case, but I'm not sure if it should be taken into account inside that func and not outside (where dispelStrength is calculated).

kazzmir: I think runemaster can be handled elsewhere outside this function

sidav: Changes done.

game/magic/game/game.go
```diff
@@ -5231,6 +5232,42 @@ func (game *Game) IsGlobalEnchantmentActive(enchantment data.Enchantment) bool {
     })
 }
 
+// Returns dispel chance against 250. 
+// TargetSpellCost may be either a base cost or an effective cost (https://masterofmagic.fandom.com/wiki/Casting_Cost#Dispelling_Magic)
+// If checkTargetSpellOwnerRetorts is true, then Archmage and Mastery retorts will be taken into consideration for dispel resistance
+// FIXME: add Runemaster check here (it should increase the dispel strength)
+func (game *Game) CalcDispelChance(dispelStrength int, targetSpellCost int, targetSpellRealm data.MagicType, targetSpellOwner *playerlib.Player, checkTargetSpellOwnerRetorts bool) int {
```
kazzmir, line 5239: Calc is an awkward name, how about Compute

game/magic/spellbook/data.go
```diff
@@ -69,6 +74,24 @@ func (spell Spell) Cost(overland bool) int {
     return spell.CastCost
 }
 
+func (spell Spell) SpentAdditionalCost(overland bool) int {
+    if overland {
+        switch spell.Eligibility {
+            case EligibilityBoth: return spell.CastCost * 5
```
kazzmir, line 80: Don't these cases need to take the overridecost into account

game/magic/game/game.go
kazzmir, line 5239: Also can you add some test cases in cast_test.go for this function

kazzmir, line 5239: Is the check retorts argument ever going to be false?

sidav: Done.

game/magic/spellbook/data.go
sidav: It's my bad, forgot to change returns. Done.

game/magic/game/game.go
sidav: I'm too tired to notice such things myself... Done.

game/magic/game/cast.go
```diff
@@ -50,6 +50,21 @@ const (
 func (game *Game) doCastSpell(player *playerlib.Player, spell spellbook.Spell) {
     // FIXME: if the player is AI then invoke some callback that the AI will use to select targets instead of using the GameEventSelectLocationForSpell
 
+    // Tranquility effect: if it's a chaos spell, it should either resist a strength 500 dispel check or fizzle right away.
+    if spell.IsOfRealm(data.ChaosMagic) {
+        for _, checkingPlayer := range game.Players {
+            // FIXME: Not sure if multiple instances of Tranquility stack or are checked separately.
+            if checkingPlayer != player && player.GlobalEnchantments.Contains(data.EnchantmentTranquility) {
```
kazzmir, line 57: should this be `checkingPlayer.GlobalEnchantaments.Contains()` because the player casting the spell is `player`, so the other player that owns tranquility is `checkingPlayer`

kazzmir:
its probably also a good idea to add a new scenario into test/overworld where an enemy has tranquility enabled, and the human player can try to cast various chaos spells to see if they fizzle due to the tranquility

kazzmir:
looking good so far

kazzmir (approved):

PR #371, opened 2025-02-23
cast: add unrest from Armageddon and Great Wasting

msom:
Armageddon and Great Wasting also adds unrest. I went with the [Code from 1.31](https://github.com/jbalcomb/ReMoM/blob/main/src/CITYCALC.C#L3282) + the indicated fixes:
- Armageddon adds +2 unrest per enemy enchantment
- Great Wasting adds +1 unrest per enemy enchtantment

I also add protection by Consecration and Chaos Ward for the city centers.

What do you think, does that make sense @kazzmir @sidav?

### DIFF
--- a/game/magic/game/game.go
+++ b/game/magic/game/game.go
@@ -6775,9 +6775,16 @@ func (game *Game) doArmageddon() {
                     for y := range mapObject.Map.Rows() {
                         point := data.PlanePoint{X: x, Y: y, Plane: mapObject.Plane}
                         tile := terrain.GetTile(mapObject.Map.Terrain[x][y])
-                        if !tile.IsWater() && !tile.IsRiver() && !mapObject.HasVolcano(x, y) && !mapObject.HasMagicNode(x, y) && !catchment.Contains(point) {
-                            points = append(points, point)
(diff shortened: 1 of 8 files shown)


game/magic/game/game.go
```diff
@@ -7797,3 +7811,11 @@ func (game *Game) DrawGame(screen *ebiten.Image){
 
     game.HudUI.Draw(game.HudUI, screen)
 }
+
+func (game *Game) GetAllGlobalEnchantments() map[data.BannerType]set.Set[data.Enchantment] {
+    enchantments := make(map[data.BannerType]set.Set[data.Enchantment])
+    for _, player := range game.Players {
+        enchantments[player.GetBanner()] = *player.GlobalEnchantments
```
kazzmir, line 7818: this is slightly dangerous in that if the underlying set is mutated then that will impact the player.GlobalEnchantments because `set1 = set2` will make a copy by reference of the map in the set. Could we use a deep clone of the set here to be safe?
```
enchantments[player.GetBanner()] = player.GlobalEnchantments.Clone()
```
which would imply the map type of enchantments would have to be `map[data.BannerType]*set.Set`

kazzmir:
I made doArmageddon and doGreatWasting cache the city positions by using game.ComputeCityStackInfo()

Issue #376, opened 2025-02-24
Fonts with outlines and drop shadows


msom:
Some fonts have a (black) outline not covered by the palette.

This is an example with the first font. The palette colors are 1. transparent, 2. dark yellow, 3. light yellow. It seems like outline pixels are added when:
- no pixel from the second and third palette (yellows) are present
- at least one pixel from the third palette (dark yellow) is within reach of 1 pixel

<img width="689" alt="Image" src="https://github.com/user-attachments/assets/6cf9eab7-236a-4470-a3f6-25d8cec773ac" />

This is another example:

<img width="307" alt="Image" src="https://github.com/user-attachments/assets/ec42fccf-f540-44ba-bf0b-4db336f02eb1" />

The first palette color (transparent) is somewhat overlapping but is missing some of pixels and has some additional pixels (for example on the lower right of the "p":

<img width="542" alt="Image" src="https://github.com/user-attachments/assets/e71cd870-62c1-4e49-80b0-dcfcfaa82d31" />

Drop shadows are similar. Pixels are added if:
- no pixel from the second and third palette (yellows) are present
- at least on pixel from the second or third palette is on the left or above

This is an example with white / dark gray:
<img width="758" alt="Image" src="https://github.com/user-attachments/assets/97f79d06-dc3c-46a0-9d10-ed6074236172" />

Maybe it's possible to define something similar to `MakeOptimizedFontWithPalette` but which draws the glyphs with outlines / dropshadows

kazzmir:
Maybe we can use something similar to the edge shader?

msom:
That might be easier, especially since the outline/shadow is also in the spacing between glyphs...

kazzmir:
mostly resolved by #377

PR #377, opened 2025-02-24
font: draw outline around glyphs

kazzmir:
image


kazzmir:
closes https://github.com/kazzmir/master-of-magic/issues/376

PR #382, opened 2025-02-25
cast: add detect magic

msom:
Detect Magic:
- Shows the current spell in the magic view above the portrait for known enemies
- Shows a shortened version of the summoning animation (for known enemies)
- Shows the normal animation for global enchantments (for known enemies)
- Shows the normal animation for city enchantments (for known cities)

The last part with city enchantments is not implemented yet. We first need to make city enchantments castable for AI, maybe do casting using two events: first select location, then apply effect and show casting animation. The first event could be skipped for AI entirely; and in the second the casting animation for AI if not detectable by the human player.

I also move the font definitions the the new library, made the colors more similar to the original and added drop shadows. And added autocache to the fontviewer.

Bildschirmfoto 2025-02-25 um 07 08 54

### DIFF
--- a/game/magic/magicview/view.go
+++ b/game/magic/magicview/view.go
@@ -595,7 +595,6 @@ func (magic *MagicScreen) MakeUI(player *playerlib.Player, enemies []*playerlib.
         Rect: transmuteRect,
         PlaySoundLeftClick: true,
         LeftClick: func(element *uilib.UIElement){
-            // FIXME:
             transmuteGroup := MakeTransmuteElements(ui, fonts.TransmuteFont, player, &help, magic.Cache, &magic.ImageCache)
             ui.AddGroup(transmuteGroup)
         },


game/magic/magicview/view.go
```diff
@@ -615,7 +595,8 @@ func (magic *MagicScreen) MakeUI(player *playerlib.Player, enemies []*playerlib.
         Rect: transmuteRect,
         PlaySoundLeftClick: true,
         LeftClick: func(element *uilib.UIElement){
-            transmuteGroup := MakeTransmuteElements(ui, transmuteFont, player, &help, magic.Cache, &magic.ImageCache)
+            // FIXME:
```
kazzmir, line 598: whats this fixme for

kazzmir:
https://masterofmagic.fandom.com/wiki/Detect_Magic mentions showing summoning spells, but summoning spells only talks about creatures/heros, not artifacts. Is detect magic supposed to show an artifact made by an enemy wizard?

kazzmir (approved):

PR #389, opened 2025-02-26
combat: show unit info on right click

kazzmir:
image

right click to bring up the unit info popup during combat.

not sure if I will keep this, its just an experiment


kazzmir:
dos mom does show this ui, so we should keep it!

image

PR #399, opened 2025-02-28
cast: add lycanthropy

msom:
Lycanthropy
- keeps weapon bonus
- keeps damage
- resets experience
- cannot be cast on fantastic units, heroes or undead / black or chaos channeled units and will generate an error message while allowing to select another unit

I also
- added notices when trying to cast a unit or city enchantments on a unit / city which already contains the enchantment (it allows to select another unit / city)
- made black channeled units (irreversibly) undead
- changed showing a notice always playing the beep

For this PR, I have done some refactoring, mostly to allow the notices and to remove some code duplication:
- move `selected` to `doCastCityEnchantments` similar to `doCastUnitEnchantment`
- added `before` and `after` callbacks to `doCastCityEnchantments`
- added `before` and `after` callbacks to `doCastUnitEnchantment`

### DIFF
--- a/game/magic/game/cast.go
+++ b/game/magic/game/cast.go
@@ -46,6 +46,10 @@ const (
     LocationTypeDisenchant
 )
 
+type UnitEnchantmentCallback func (*units.OverworldUnit)
+
+func noUnitEnchantmentCallback(unit *units.OverworldUnit) {}
+
 type CityEnchantmentCallback func (*citylib.City) bool
 
 func noCityEnchantmentCallback(city *citylib.City) bool {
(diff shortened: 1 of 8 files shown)


kazzmir:
Works good but the sound is wrong. When I cast lycanthropy the sound played is the shuffle sound in combat.

game/magic/game/cast.go
```diff
@@ -87,79 +99,113 @@ func (game *Game) doCastSpell(player *playerlib.Player, spell spellbook.Spell) {
             UNIT ENCHANTMENTS
         */
         case "Bless":
-            game.doCastUnitEnchantment(player, spell, data.UnitEnchantmentBless)
+            game.doCastUnitEnchantment(player, spell, data.UnitEnchantmentBless, noUnitEnchantmentCallback, noUnitEnchantmentCallback)
```
kazzmir, line 102: Could there be another method like doCastUnitEnchantmentFull that takes these extra two arguments, so that the unit enchantments that don't need the extra functionality don't need to change. Then the doCastEnchantment can involve the Full method with the two extra noXyz arguments.

And same for the city enchantments

game/magic/game/cast.go
```diff
@@ -153,13 +165,47 @@ func (game *Game) doCastSpell(player *playerlib.Player, spell spellbook.Spell) {
         case "Flame Blade":
             game.doCastUnitEnchantment(player, spell, data.UnitEnchantmentFlameBlade)
         case "Black Channels":
-            game.doCastUnitEnchantment(player, spell, data.UnitEnchantmentBlackChannels)
+            after := func (unit units.StackUnit) bool {
+                overworldUnit, ok := unit.(*units.OverworldUnit)
+                if ok {
+                    overworldUnit.Undead = true
```
kazzmir, line 171: I think this can be set on a hero, https://masterofmagic.fandom.com/wiki/Black_Channels

so StackUnit should just have a SetUndead() method that works for both.

kazzmir:
> > Works good but the sound is wrong. When I cast lycanthropy the sound played is the shuffle sound in combat.
> 
> Actually it's the same sound as in original dos mom... Not sure what they intended by using this sound, maybe the sound of many werewolfs running?

Thats pretty funny.. its ok for now. I would have expected the sound to be similar to Wraith Form, sound 32

kazzmir (approved):

PR #405, opened 2025-03-01
cast: cruel unminding and drain power

sidav:


game/magic/game/cast.go
```diff
@@ -1357,6 +1359,39 @@ func (game *Game) doCastSpellBlast(player *playerlib.Player) {
...
+func (game *Game) doCastCruelUnminding(player *playerlib.Player) {
+    var wizSelectionUiGroup *uilib.UIElementGroup
+    onTargetSelectCallback := func(targetPlayer *playerlib.Player) (bool, string) {
+        if targetPlayer.Defeated || targetPlayer.Banished {
+            return false, ""
+        }
+        targetSkill := targetPlayer.ComputeCastingSkill()
+        minReduction := max(targetSkill / 100, 1)
+        maxReduction := max(targetSkill / 10, 1)
+        reduction := minReduction + rand.N(maxReduction)
```
kazzmir, line 1371: should this be `minReduction + rand.N(maxReduction - minReduction)` ? since you want a random value between min,max, so the random part is the distance between min and max

game/magic/game/cast-select-wizard.go
```diff
@@ -181,3 +181,166 @@ func makeSelectSpellBlastTargetUI(ui *uilib.UI, cache *lbx.LbxCache, imageCache
...
+            mx, my := options.GeoM.Apply(float64(84 * data.ScreenScale), float64(10 * data.ScreenScale))
+            fonts.BigOrange.PrintWrapCenter(screen, mx, my, 150. * float64(data.ScreenScale), float64(data.ScreenScale), options.ColorScale, header)
+        },
+        Order: 0,
+    })
+
+    // Wizard faces/gems/broken gems
+    crystalPicture, _ := imageCache.GetImage("magic.lbx", 6, 0)
+    brokenCrystalPicture, _ := imageCache.GetImage("magic.lbx", 51, 0)
+    wizardFacesOffsets := [][2]int {{24, 37}, {101, 37}, {24, 98}, {101, 98}}
```
kazzmir, line 240: a minor point, but this could be an array of image.Point.
```
[]image.Point{image.Pt(24, 37), image.Pt(101, 37), ...}
```
and then later on
```
x + wizardFacesOffsets[index].X, y + wizardFacesOffsets[index].Y
```
You can change it if you like, but I'll probably merge it either way

game/magic/game/cast.go
sidav: Ow, you're right. Will fix.

game/magic/game/cast-select-wizard.go
sidav: Done.

game/magic/game/cast.go
```diff
@@ -1357,6 +1359,40 @@ func (game *Game) doCastSpellBlast(player *playerlib.Player) {
...
+        minReduction := max(targetSkill / 100, 1)
+        maxReduction := max(targetSkill / 10, 1)
+        reductionRandomSpread := max(maxReduction - minReduction, 1) // It should never be zero, or rand.N will crash
+        reduction := minReduction + rand.N(reductionRandomSpread)
+        actuallyReduced := targetPlayer.ReduceCastingSkill(reduction)
+        return true, fmt.Sprintf("%s loses %d points of casting ability", targetPlayer.Wizard.Name, actuallyReduced)
+    }
+    playersInGame := len(game.Players)
+    wizSelectionUiGroup = makeSelectTargetWizardUI(game.HudUI, game.Cache, &game.ImageCache, "Select target for Cruel Unminding spell", 41, player, playersInGame, onTargetSelectCallback)
+    game.HudUI.AddGroup(wizSelectionUiGroup)
```
kazzmir, line 1378: To restate what I mentioned in discord: the next line should run a yield loop until the user has finished interacting with the wizSelectionUIGroup. That loop will look like
```
for !quit {
  game.Counter += 1 // to let animations play in the background
  game.HudUI.StandardUpdate() // to let the ui be usable
  yield()
```
Where `!quit` is some condition that should become false once the user has selected a wizard or pressed the cancel button. That condition can be set in any number of ways, but one easy way is to use a context.
```
quit, cancel := context.WithCancel(context.Background())
wizSelectionUiGroup := makeSelectTargetWizardUI(..., cancel)
for quit.Err() == nil {
  ...
  yield()
}
```
Where the ui will invoke the cancel function at the place where the group is removed, `ui.RemoveGroup(group)`

kazzmir:
If the only changes left are the sounds, we could merge this PR as-is and fix the sounds in a future PR

kazzmir (approved):

PR #406, opened 2025-03-01
city: implement unit reduction bonus

msom:
Coal and iron do not increase production but [reduce the production cost of most normal units](https://masterofmagic.fandom.com/wiki/Iron_Ore#Unit_Cost_Reduction).


game/magic/cityview/build-screen.go
```diff
@@ -243,7 +243,7 @@ func makeBuildUI(cache *lbx.LbxCache, imageCache *util.ImageCache, city *citylib
 
                 options.GeoM.Reset()
                 options.GeoM.Translate(float64(130 * data.ScreenScale), float64(7 * data.ScreenScale))
-                unitview.RenderUnitInfoBuild(screen, imageCache, bannerUnit, fonts.DescriptionFont, fonts.SmallFont, options)
+                unitview.RenderUnitInfoBuild(screen, imageCache, bannerUnit, fonts.DescriptionFont, fonts.SmallFont, options, city.UnitProductionCost(&unit))
```
kazzmir, line 246: it might actually be good to cache the production cost in a variable outside the call to AddElement() so that this is not recomputed each draw call

game/magic/city/city.go
```diff
@@ -1544,6 +1532,30 @@ func (city *City) WorkProductionRate() float32 {
...
+
+    if city.Race == data.RaceDwarf {
+        reduction *= 2
+    }
+
+    if city.Buildings.Contains(buildinglib.BuildingMinersGuild) {
+        reduction *= 2
+    }
+
+    return unit.ProductionCost - int(float32(unit.ProductionCost) * reduction)
```
kazzmir, line 1556: The wiki says 'The maximum possible bonus is -50%'. Should this use `min(reduction, 0.5)`

kazzmir (approved):

PR #411, opened 2025-03-02
cast: add eternal night

msom:
Adds the [Eternal Night](https://masterofmagic.fandom.com/wiki/Eternal_Night) global enchantment, which adds Darkness to battles if not already under the influence of Cloud of Shadow.

Not 100% sure what the wiki means with "protected" here:  "every battle that does not take place in a Town protected by Heavenly Light comes under an effect virtually identical to that of the Darkness spell". Since True Light negates Darkness and Eternal Night just adds Darkness I guess we should just add it and let Heavenly Light "protect" by negating the effects.

I also added a check to `AddEnchantment` since DOS MoM does not allow casting Darkness/True Light if is already preset as global enchantment.

### DIFF
--- a/game/magic/combat/combat-screen.go
+++ b/game/magic/combat/combat-screen.go
@@ -1659,7 +1659,7 @@ func (combat *CombatScreen) MakeUI(player *playerlib.Player) *uilib.UI {
 
     // info
     elements = append(elements, makeButton(20, 0, 1, func(){
-        // FIXME
+        // FIXME: show enchantments such as "Eternal Night", "Cloud of Shadow", "Heavenly Light" etc.
     }))
 
     // auto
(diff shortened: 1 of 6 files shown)


kazzmir:
I had put enchantments that are the result of some external reason (such as heavenly light or cloud of darkness) into a separate GlobalEnchantments array in the CombatModel because initially I didn't think those enchanments should show up on either side. I guess the combat model could add true light/darkness to the appropriate side (whoever owns the global enchantment), but I agree its unclear where to put the enchantment if neither side owns the global enchantment.

kazzmir:
Part of the reason the global enchantments exist in the combat model is to prevent the player from dispelling those enchantments, since any enchantment that shows up in the UI under the wizard's name can be clicked on to dispel.

kazzmir (approved):

PR #417, opened 2025-03-03
mapeditor: allow loading from savegame

msom:
@kazzmir what do you think about allowing to load a map from an existing savegame?

### DIFF
--- a/util/mapeditor/main.go
+++ b/util/mapeditor/main.go
@@ -5,8 +5,10 @@ import (
     "fmt"
     "log"
     "time"
+    "flag"
     "image/color"
     "math/rand"
+    "encoding/binary"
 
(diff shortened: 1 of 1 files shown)


util/mapeditor/main.go
```diff
@@ -320,6 +322,42 @@ func MakeEditor() *Editor {
...
+    }
+    defer reader.Close()
+
+    _, err = reader.Seek(fileOffset, 0)
+    if err != nil {
+        fmt.Printf("Error seeking file: %v", err)
+        os.Exit(0)
+    }
+
+    editor.Map = terrain.MakeMap(40, 60)
```
kazzmir, line 347: are game worlds always 40x60? even if they are small/medium/large?

kazzmir:
this is great!

kazzmir (approved):

Issue #335, opened 2025-02-18
combat song plays note indefinitely


kazzmir:
Sometimes SongCombat1 will play a note for too long so the note is held until the game exits, because the notes don't seem to get cut off.

kazzmir:
I think this mostly occurs in the browser, potentially because due to some cpu lagging the midi note doesn't get sent properly

kazzmir:
closed due to https://github.com/kazzmir/master-of-magic/pull/520

PR #520, opened 2025-04-26
midi: embed pure golang midi synthesizer

kazzmir:
Use a built-in midi synthesizer instead of a c++ external driver + fluidsynth. This will search for .sf2 files in the embedded data/ folder, as well as in the working directory of the executable and in /usr/share/sounds, which is where fluidsynth sf2 files are normally installed. The timidity 6mb sf2 file is included in data/data so that there is some sound at least, but fluidsynth_gm2.sf2 sounds the best (but it is 140mb)


kazzmir:
closes https://github.com/kazzmir/master-of-magic/issues/335

Issue #350, opened 2025-02-19
Implement proper spellbook page behavior


sidav:
Currently, the spellbook is always opened on the first page.
The behavior from the original game is the following:

1. SPELL button: 
  - If your spell is being cast right now, the book always opens on the page containing the spell.
  - If nothing is being cast right now, the last opened page is used.
2. Info -> Apprentice:
  - It is always opened on the page with ongoing research.
  - Behavior when all the spells are researched is to be determined

PR #532, opened 2025-05-12
spellbook: track which spell book page the user had opened

kazzmir:
fixes https://github.com/kazzmir/master-of-magic/issues/350

The apprentice spell book opens to the currently researching spell, or just to the first page of the book if there are no spells being researched (because the wizard has researched them all).

The combat spell book will open to the last page that was shown.


kazzmir:
closes #350

PR #701, opened 2026-01-13
ai: more improvements

kazzmir:


gabe647:
This branch now conflicts with master. The still-applicable Enemy2 bits (to-hit in power, always-build, tax/alchemy, ConfirmEncounter, quieter logs) are rebased in #740.

Left out the watch-mode “no raiders / 1 opponent” change — debug default, not AI logic. #739’s lair planner will replace the simple ConfirmEncounter check if that stack lands.

PR #710, opened 2026-04-08
Fix issues about SetRoad

DinoHuang-git:
1. Fix crash when SetRoad: Initial map first
2. SetRoad when create outpost: Cities should have roads by default.
3. Fix crash in MaybeBuildRoads: Fix crash if try to build road in a city


kazzmir:
On Myrror when a unit moves into a newly built city the unit uses up 0 movement points, as if that city had a road there. On the other hand, a city with no roads around it that gets razed does not leave a road on the map. For now I'll go with always creating a road in the city when it is created. I may revisit this later on when adding roads in combat in town battles.

PR #715, opened 2026-06-03
combat: move magic vortex automatically for all AI armies

alexfibreplanner:
In watch mode the player's army is AI-controlled but not flagged Auto, so the magic-vortex code prompted for a manual placement that never came and combat stalled.


kazzmir:
This is fine although there are no players armies in play, only the AI controlled wizards actually do anything. This must mean that a magic vortex owned by an ai player would always require input from the human

kazzmir (approved):

PR #720, opened 2026-08-14
Add 'End Of Turn Wait' setting

gabe647:
## Summary
- Adds the original game's "End Of Turn Wait" toggle to the settings screen (title screen and in-game).
- When disabled, the human player's turn automatically ends once every stack is out of moves and nothing is selected — no need to click Next Turn manually.
- Defaults to enabled (current wait-for-click behavior), so this is non-breaking.

## Implementation notes
- `Player.AllStacksOutOfMoves()` (player/player.go) — new helper, reuses the existing `UnitStack.OutOfMoves()` per-stack check (which already correctly excludes patrolling/busy units).
- The setting itself lives on `Music` rather than `Game`, since `Music` is the one object already shared between the main menu and an in-progress game — same reasoning as why volume lives there. Mirrors the existing `GetVolume`/`SetVolume` pattern with `GetEndOfTurnWait`/`SetEndOfTurnWait`.
- The auto-advance check is added inside `Game.Update()`, gated behind the same "no UI modal open" condition that already gates player input, so it can't fire while a dialog (e.g. the disband-confirmation prompt) is open.
- Settings UI checkbox is hand-drawn (vector.FillRect/DrawRect) since there was no existing checkbox component in `uilib` to reuse.

## Test plan
- [x] `go build` (native) and `go vet` clean
- [x] `go test ./...` passes
- [x] Native build: verified the checkbox renders checked by default, toggles on click, and unchecking it causes turns to auto-advance once all stacks are out of moves
- [x] WASM build: verified the settings screen (including the new checkbox) renders correctly in-browser

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,18 +1,23 @@
 # macOS system files
 .DS_Store
 
+# Personal Claude Code settings (not team-shared)
+.claude/settings.local.json
+
 # Visual Studio Code editor settings
 .vscode/
 
 # Data files
 data/data/*.zip
 data/data/*.LBX
+/data/mom/
 
 # Executables
 /lbxdump
 /lbxviewer
 /magic
 /mapeditor
+*.wasm
 
 # Memory and CPU profiles
 profile.cpu.*
--- a/game/magic/game/game.go
+++ b/game/magic/game/game.go
@@ -1662,7 +1662,8 @@ type SettingsUI struct {
 }
 
 func (settings *SettingsUI) RunSettingsUI() {
-    group, quit := settingslib.MakeSettingsUI(settings.Game.Cache, &settings.Game.ImageCache, settings.Game.Music)
+    game := settings.Game
+    group, quit := settingslib.MakeSettingsUI(game.Cache, &game.ImageCache, game.Music, game.Music.GetEndOfTurnWait, game.Music.SetEndOfTurnWait)
     settings.Game.doRunUI(settings.Yield, group, quit)
 }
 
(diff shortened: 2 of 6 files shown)


kazzmir:
this is probably ok, ill merge it after a bit more review/testing

kazzmir:
image

can the left side of the box for end of turn be vertically aligned with the left side of the box for volume.

also probably make the text 'end of turn wait' to be aligned to the middle of the checkbox

gabe647:
Pushed a fix for the alignment feedback:

- Checkbox's left edge now matches the volume box's left edge (both at x=29, accounting for the volume box's 1px inset).
- "End Of Turn Wait" label is now vertically centered on the checkbox, computed from the font's actual glyph height instead of a hardcoded offset.

Verified in a native run of the settings screen — checkbox and volume box left edges line up, and the label sits centered. Screenshot attached in chat; happy to add it here too if useful.

gabe647:
Updated screenshot after also centering the label text vertically on the checkbox (the font's glyph cell reserves descender space this string doesn't use, so centering on cell height alone read as slightly too high):

end-of-turn-wait-settings-final

kazzmir:
Ok cool. I think storing this info in the Music object is probably not right. Maybe there can be some kind of Settings object. I'll think about it a little bit.

> Screenshot attached in chat; happy to add it here too if useful.

btw just so I don't misunderstand, by 'chat' do you mean this github conversation? or some other thing?

gabe647:
> Ok cool. I think storing this info in the Music object is probably not right. Maybe there can be some kind of Settings object. I'll think about it a little bit.
> 
> > Screenshot attached in chat; happy to add it here too if useful.
> 
> btw just so I don't misunderstand, by 'chat' do you mean this github conversation? or some other thing?

Yeah just opened another PR to refactor Settings as a first-class object.  Chat was just the PR comments here.

kazzmir:
This is superseded by #724

PR #722, opened 2026-08-14
Add 'Strategic Combat Only' setting

gabe647:
## Summary
- Adds the original game's "Strategic Combat Only" toggle to the settings screen (title screen and in-game).
- When enabled, the human player's own battles are auto-resolved via the existing `combat.DoStrategicCombat` instead of playing out on the battle screen.
- Defaults to disabled (current always-show-battle-screen behavior for the human), so this is non-breaking.

## Implementation notes
- `Player.StrategicCombat` and `combat.DoStrategicCombat` already existed and were already used for AI-vs-AI encounters (always strategic) and wild-monster encounters, but were never reachable by the human player - this setting is what was missing to wire it up.
- `doCombat` (game/game.go) now checks `attacker.StrategicCombat && defender.StrategicCombat` (only meaningful when a human is involved and not in watch mode) and skips straight to `DoStrategicCombat`, taking its returned state/casualty counts directly instead of reading them from the unused `combatModel`. Existing AI-vs-AI behavior via `combat.Run` is untouched - that path is only reached when no human is involved, so it can't be affected by this human-facing preference.
- Same pattern as End Of Turn Wait (#720): the setting lives on `Music` since it's shared between the main menu and an in-progress game. Toggling it in-game also applies immediately to the current human player (not just future games), via `RunSettingsUI`'s setter.

## Test plan
- [x] Verified on the native build: checkbox renders/toggles correctly, and with it enabled a real battle (spearmen vs. stronger opposition) was auto-resolved and lost as expected instead of opening the battle screen

### DIFF
--- a/game/magic/game/game.go
+++ b/game/magic/game/game.go
@@ -620,6 +620,10 @@ func MakeGameWithModel(lbxCache *lbx.LbxCache, music_ *music.Music, makeModel fu
     // game.Model = MakeGameModel(terrainData, settings, data.PlaneArcanus, game.Events, heroNames, game.AllSpells(), createArtifactPool(lbxCache), buildingInfo)
     game.Model = makeModel(lbxCache, game.Events)
 
+    if len(game.Model.Players) > 0 {
+        game.Model.GetHumanPlayer().StrategicCombat = game.Music.GetStrategicCombatOnly()
+    }
+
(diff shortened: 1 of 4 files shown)


kazzmir:
:+1:

kazzmir:
This is superseded by #724

PR #723, opened 2026-08-14
Implement Spell of Mastery win condition

gabe647:
## Summary

Casting Spell of Mastery to completion currently falls through to `doCastSpell`'s `default:` case ("Warning: casting unhandled spell") - there's no win condition anywhere in the codebase. This wires it up.

The `mastery` package already had everything needed for the presentation - `LabVortexScreen` (banishes the other wizards) and `SpellOfMasteryEndScreen` (the "Master of Magic" victory screen, using `win.lbx`) - but neither was ever called from anywhere. Only `ShowSpellOfMasteryScreen` (the "X has started casting" announcement, shown when casting *begins*) was wired up.

## Implementation

- `doCastSpell` (cast.go) gets a `case "Spell of Mastery"`. It can't run the victory screens directly - it has no `yield` (it's called from `StartPlayerTurn`'s non-yielding per-turn casting-progress code, among other places) - so it queues a new `GameEventSpellOfMasteryComplete` event instead.
- `ProcessEvents` (game.go) handles that event by calling the new `doSpellOfMasteryVictory(yield, player)`, which runs the vortex screen then the end screen (reusing the existing `PushDrawer`/coroutine pattern already used for the casting-announcement screen), then sets `GameStateQuit` to return to the main menu - no new `GameState` needed, this reuses the existing quit-to-menu path.
- Any other events queued during the same `StartPlayerTurn` pass (e.g. a research-spell prompt, since research and casting progress are both processed there) are drained right after, otherwise `ProcessEvents`'s drain loop keeps handling them regardless of the state change and the player ends up back in a live turn instead of at the menu. Caught via manual testing, see below.
- Applies uniformly whether the human or an AI wizard completes the cast, matching the original game (the human sees the winning wizard's own face/portrait either way).

## Test plan

- [x] Manual playtest via a save-game fixture (built with a temporary local dev flag, not part of this diff) with the human player one mana point from completing Spell of Mastery: advancing one turn triggers the win, both screens play, and it lands cleanly back at the main menu with no leftover UI prompts (this last part needed the event-draining fix above - without it, a queued research-spell prompt would still get shown after the win screens, and only dismissing that returned you to the menu)

### DIFF
--- a/game/magic/game/cast.go
+++ b/game/magic/game/cast.go
@@ -424,9 +424,17 @@ func (game *Game) doCastSpell(player *playerlib.Player, spell spellbook.Spell) {
 
         /*
             INSTANT SPELLS
-                TODO:
-                Spell of Mastery
         */
+        case "Spell of Mastery":
+            // the casting-started announcement already played when the player chose to
+            // start casting this spell (see game.go). completion of casting is the win
+            // condition, so queue an event rather than run the victory screens directly -
+            // this function does not have a yield to drive them (and it can be invoked
+            // from the non-yielding per-turn casting-progress code path).
+            select {
+                case game.Events <- &GameEventSpellOfMasteryComplete{Player: player}:
+                default:
+            }
         case "Spell of Return":
             selected := func (yield coroutine.YieldFunc, tileX int, tileY int){
                 city := player.FindCity(tileX, tileY, game.Model.Plane)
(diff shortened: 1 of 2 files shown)


kazzmir:
Seems to work. Another easy way to test this kind of thing is to run the overworld test `go run ./test/overworld` and then press the tilda key to bring down a quake-like console, then type in 'cast Spell of mastery' and it will just auto-cast the spell for you.

I think maybe the ending animation should show the two worlds spinning above the wizard's hand indefinitely until the user clicks a mouse button. I'd have to check what dos mom does, but right now the animation just kind of ends abruptly lol

kazzmir (approved):

PR #724, opened 2026-08-14
Add rebindable keybindings and a Keys settings screen

gabe647:
## Summary

Adds a rebindable keybinding system and the missing "Keys" button/screen in Settings (previously just an empty frame slot in the background art, alongside "Ok").

**Settings refactor (prerequisite):** introduces a top-level `Settings` object (`game/magic/settings`) holding `Music`, `EndOfTurnWait`, `StrategicCombatOnly`, and the new `Keybindings`, threaded through `MagicGame`/`Game`/`MainScreen` in place of a bare `*music.Music`. `Music` was never really the right home for `EndOfTurnWait`/`StrategicCombatOnly` (added in #720/#722) - it was just the one object already shared between the main menu and an in-progress game. Didn't want to repeat that a third time for ~20 keybindings.

Note: this overlaps with the still-open #720 and #722 (re-adds those two settings onto the new `Settings` struct rather than rebasing those branches) - happy to help reconcile whichever merges first/last.

**Keybindings (`game/magic/keybinds`):** an `Action` enum for the 17 hotkeys that already have a working handler today (currently reachable only by mouse, a hardcoded key, or both) - Surveyor/Cartographer/Apprentice/Historian/Astrologer/Chancellor/Tax Collector/Grand Vizier/Mirror/Next Turn (previously hardcoded F1-F9/N), plus Game/Spellbook/Armies/Cities/Magic screens, Advisors, and Switch Planes (previously mouse-only, no key at all). Each defaults to the original game's binding (G/S/A/`---`/M/I/P/F1-F9/N). The HUD's `HandleKeys` switch (game.go) now checks these bindings instead of literal `ebiten.Key` constants.

Deliberately out of scope: Quick Save, Quick Load, Quit Without Saving, and Default Item Editor aren't included - they don't exist as features at all yet (e.g. the quit button's callback is currently a no-op), so there's nothing to bind a key to.

**Keys screen:** a scrollable-ish (fits without scrolling - 17 actions across two columns) list of every action and its bound key, click a row to open a "press the key you wish to bind" modal.

## Implementation notes / things that weren't obvious

- The Keys screen and rebind modal are built as `uilib.UIElementGroup`s added to the *caller's* existing UI (mirroring how the settings screen itself is shown), not as a separate standalone `*uilib.UI` - this package has no equivalent to `Game.PushDrawer` to get a standalone UI's `Draw` actually invoked each frame. First attempt at this silently hung (group added, loop running, nothing ever drawn).
- The nested group needs an initial `yield()` before its first `StandardUpdate()` (matching `doRunUI`'s existing pattern) - otherwise the same physical click that opened it is immediately reprocessed against the new group in the same frame, which triggered an element that happened to occupy the same screen position.
- Rebindable cases must be checked *before* the fixed Tab/Shift/Space cases in the HUD's key switch, since Go's `switch` matches whichever `case` is listed first - a key rebound onto Space, for example, was previously always shadowed by the hardcoded Space case.
- The Next Turn button's brief "pressed" flash (previously only shown on a mouse click) is now shared with the keyboard path via a short frame-counter deadline, so triggering it by key looks the same as clicking it.

## Test plan

- [x] `go build`, `go vet`, `go test ./...` clean across the whole module (including `test/*` dev tools and `util/*`, which also construct `*Game`/`*MainScreen` and needed the same signature updates)
- [x] Manual playtest on the native build: Keys button opens the list from both the title screen and in-game settings, list shows correct default bindings, rebinding a row (tested Next Turn -> Space) updates the row and takes effect in actual gameplay, old binding stops working, Escape cancels a pending rebind without changing it, the six new keyboard shortcuts (G/S/A/C/M/I/P) work from the HUD

### DIFF
--- a/game/magic/game/cast.go
+++ b/game/magic/game/cast.go
@@ -2587,8 +2587,8 @@ func (game *Game) doCastGlobalEnchantment(yield coroutine.YieldFunc, player *pla
     }
 
     if song != music.SongNone {
-        game.Settings.Music.PushSong(song)
-        defer game.Settings.Music.PopSong()
+        game.Music.PushSong(song)
+        defer game.Music.PopSong()
     }
 
     fonts := MakeGlobalEnchantmentFonts(game.Cache)
--- a/game/magic/game/cast.go
+++ b/game/magic/game/cast.go
@@ -2587,8 +2587,8 @@ func (game *Game) doCastGlobalEnchantment(yield coroutine.YieldFunc, player *pla
     }
 
     if song != music.SongNone {
-        game.Music.PushSong(song)
-        defer game.Music.PopSong()
+        game.Settings.Music.PushSong(song)
+        defer game.Settings.Music.PopSong()
     }
 
     fonts := MakeGlobalEnchantmentFonts(game.Cache)
--- a/game/magic/game/game.go
+++ b/game/magic/game/game.go
@@ -3388,6 +3388,8 @@ func (game *Game) doMoveSelectedUnit(yield coroutine.YieldFunc, player *playerli
         return
     }
 
+    player.MovedStacksThisTurn += 1
+
     mapUse := game.GetMap(stack.Plane())
 
     stepsTaken := 0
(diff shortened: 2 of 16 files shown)


kazzmir:
I agree with this general direction (I mentioned needing some kind of Settings object in #720) but I don't think Music should be directly a part of Settings. I'll look more into this in a bit

kazzmir:
that test was just to see if github would let me push to your branch, seems like it works, so ill just make edits here

gabe647:
> I agree with this general direction (I mentioned needing some kind of Settings object in #720) but I don't think Music should be directly a part of Settings. I'll look more into this in a bit

Music could be a separate top-level dep, I thought putting it as a child of Settings might be cleaner but it's only referenced in a few places so either approach should work.

kazzmir:
Ok I moved Music to be a property of MagicGame and then it just gets passed around. Also I fixed an issue where if end of next turn was disabled and the player has no units active (all dead or on patrol) then the game will require the user to click 'next turn' manually, otherwise the game would just fly by without the user being able to do anything at all. I verified this mechanic on dos mom.

kazzmir (approved):

gabe647 (commented):
Tested locally, working as expected

PR #726, opened 2026-08-15
Fix infinite disband-warning loop with auto next turn

gabe647:
## Summary
- With "end of turn wait" disabled, the auto-advance check in `Update()` (game.go) fires `GameEventNextTurn` every frame once the human player has no stack selected and no stacks with moves left.
- `doNextTurn`'s food/gold/mana disband confirmation doesn't change any of that state when the player clicks "No" - so on the very next frame the auto-check's condition was still true and it immediately re-fired `GameEventNextTurn`, reshowing the same dialog. From the player's perspective, clicking "No" appeared to do nothing.
- Found via manual playtesting after the "end of turn wait" / keybindings work in #724 landed.

## Fix
- Adds `Player.SuppressAutoNextTurn`, set `true` when the disband warning is declined.
- The auto-advance condition additionally requires `!player.SuppressAutoNextTurn`.
- Reset back to `false` in `StartPlayerTurn`, alongside the existing `MovedStacksThisTurn` reset, so it only suppresses for the remainder of the current turn.

## Test plan
- [x] Verified on the native build: with End Of Turn Wait disabled, reached a turn with a food shortage, clicked "No" on the warning - now correctly returns control to the map instead of immediately reshowing the same dialog

### DIFF
--- a/game/magic/game/game.go
+++ b/game/magic/game/game.go
@@ -2181,6 +2181,7 @@ func (game *Game) doNextTurn(yield coroutine.YieldFunc) {
 
         no := func(){
             quit = true
+            player.SuppressAutoNextTurn = true
             game.HudUI.RemoveGroup(group)
         }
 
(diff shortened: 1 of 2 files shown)


kazzmir:
I added you as a collaborator to this repo, if you want to remake this PR directly off this repo. At least then the github ci checks will run

gabe647:
Superseded by #727 (same-repo branch so CI actually runs against it).

PR #728, opened 2026-08-15
Add 'Background Music' setting

gabe647:
## Summary
- Wires the existing `Music.Enabled` flag up to a settings screen checkbox. `PlaySong` already checks it live (`music/music.go`), it was previously only settable via the `-music` CLI flag at startup.
- Renamed the `settings` package's `VolumeSettings` interface to `MusicSettings` and added `IsMusicEnabled`/`SetMusicEnabled` to `Music`, keeping `settings.go` decoupled from the `music` package (same pattern as the existing `GetVolume`/`SetVolume`).
- Toggling off calls `Stop()` immediately. Toggling back on resumes whatever song context is on top of the existing song-queue stack, rather than staying silent until some unrelated part of the game happens to call `PlaySong` next.

## Scope note
This maps to the single existing on/off flag - i.e. all music. The original game's separate "Event Music" checkbox would need songs classified into background vs. event categories, which doesn't exist in this codebase yet, so it's left as a separate future addition rather than attempted here.

## Test plan
- [x] Verified on the native build: checkbox defaults checked, unchecking stops music immediately, rechecking resumes music immediately (caught and fixed a bug during testing where re-enabling only flipped the flag without resuming playback)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,12 +7,14 @@
 # Data files
 data/data/*.zip
 data/data/*.LBX
+/data/mom/
 
 # Executables
 /lbxdump
 /lbxviewer
 /magic
 /mapeditor
+*.wasm
 
 # Memory and CPU profiles
 profile.cpu.*
--- a/game/magic/music/music.go
+++ b/game/magic/music/music.go
@@ -220,6 +220,22 @@ func (music *Music) SetVolume(volume float64) {
     }
 }
 
+func (music *Music) IsMusicEnabled() bool {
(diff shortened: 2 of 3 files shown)


kazzmir:
lol so many PRs. is ai helping you :)

gabe647:
> lol so many PRs. is ai helping you :)

I don't know if it's helping per se, but it's definitely doing something :)

kazzmir (approved):

PR #733, opened 2026-08-15
artifact: dedup item-ability bitmask between create-artifact and item

gabe647:
# What
Pure refactor: eliminate a duplicated table.

`ReadPowers` in `create-artifact.go` built a 32-entry `map[uint32]data.ItemAbility`
(bit → item ability) that is byte-for-byte identical to the one returned by
`GetItemConversionMaps()` in `item.go`. This reuses the canonical maps instead of
re-declaring the copy.

Both functions parse the same LBX data (`itempow.lbx` / `itemdata.lbx`), so the two
tables were always in lock-step; sharing one definition means artifact parsing can no
longer silently drift from item loading.

# Details
- Removes the 32-entry literal from `ReadPowers`.
- Replaces it with `_, _, abilityMap := GetItemConversionMaps()`.
- No new exported API, no new dependencies, no runtime/behaviour change — the values
  are identical, so parsing produces the exact same results.

# Verification
- `go build ./game/magic/...` — clean.
- `go test ./game/magic/artifact/...` — pass.
- Confirmed the two removed/kept tables are byte-for-byte identical prior to the change.

### DIFF
--- a/game/magic/artifact/create-artifact.go
+++ b/game/magic/artifact/create-artifact.go
@@ -100,40 +100,9 @@ func ReadPowers(cache *lbx.LbxCache) ([]Power, map[Power]int, map[Power]set.Set[
         4: data.DeathMagic,
     }
 
-    abilityMap := map[uint32]data.ItemAbility{
-        1 << 0:  data.ItemAbilityVampiric,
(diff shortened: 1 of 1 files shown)


kazzmir:
👍

kazzmir (approved):

PR #734, opened 2026-08-15
keybinds: prevent duplicate key bindings in the Keys screen

gabe647:
## Summary

Rebinding a key in the **Keys** settings screen could silently leave two actions sharing one key, so a single keypress would fire whichever action Go's non-fall-through `switch` happened to match first.

- Add `Keybindings.ConflictingActionForKey`, which reports the action that currently owns a given key.
- Wire it into the Keys screen's rebind flow: when the chosen key is already held by another action, the previous owner is unbound first (matching the original game's behavior).
- The confirm step pops the standard confirm dialog, rendered on a new `confirmLayer` above the Keys screen, with its elements added to the same group that drives their fade/removal so the dialog shows and the button clicks cancel the nested UI pump.
- Adds unit tests for `ConflictingActionForKey` across the conflict, unbound, free-key, and rebind-tracking cases.

## Motivation

This completes the "No duplicate-key-binding detection" item flagged in #724. The Keys screen now makes it impossible for two actions to end up on the same key.

## Testing

- `go test ./game/magic/keybinds/...` passes.
- Native (`go build -o magic ./game/magic`) and WASM (`GOOS=js GOARCH=wasm go build`) both build.
- Manually verified: entering Settings → Keys, rebinding an action onto an already-used key now shows the confirm dialog; confirming unbinds the previous owner and binds the key to the new action.

## Related

Closes the duplicate-key-detection gap from #724 (rebindable keybindings + Keys screen).

### DIFF
--- a/game/magic/keybinds/keybinds.go
+++ b/game/magic/keybinds/keybinds.go
@@ -129,6 +129,24 @@ func (keybindings *Keybindings) Get(action Action) ebiten.Key {
     return keybindings.bindings[action]
 }
 
+// ConflictingActionForKey returns the action currently bound to key, or the
+// zero-value ActionGameScreen and false if no action holds that key. It is the
+// primitive the Keys screen consults before applying a rebind so a key can
+// never end up driving two actions at once.
+func (keybindings *Keybindings) ConflictingActionForKey(key ebiten.Key) (Action, bool) {
+    if key == Unbound {
+        return ActionGameScreen, false
+    }
+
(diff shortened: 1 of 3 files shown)


gabe647:
Example dialog:
Screenshot 2026-08-15 at 3 56 13 PM

kazzmir:
BTW maybe you can add a 'reset to defaults' button as well? A separate PR is fine

kazzmir (approved):

PR #737, opened 2026-08-16
artifact: add in-game Default Item Editor

gabe647:
## Summary
Community-patch-style **Default Item Editor**: an in-game replacement for
ITEMMAKE.EXE. The overland action opens the Enchant Item / Create Artifact
UI plus **Item #N** arrows so you can browse and edit the 250-slot premade
catalog.

- `artifact.Catalog` is now the source of truth for premade items: a
  250-slot slice plus an availability mask (identity is the slot index,
  not the name). Treasure, merchants, gifts, and DOS-save import award by
  index; already-awarded slots stay awarded even if you rewrite them.
- `ShowDefaultItemEditor` reuses the create-artifact helpers. No wizard
  book filter, raw item cost (not spell cost), OK writes the current slot
  and closes. Browsing Item # commits the current slot first.
- Realm requirements are written by color (Insecticide), not official
  ITEMMAKE's power-slot order.
- `keybinds`: new `ActionDefaultItemEditor`, shown on the Keys screen,
  default **E** (unused by the original A/G/I/M/N/P/S + F1–F9 set; C left
  free for Cities, F10 left for a future Quick Save). The Community Patch
  left this action unbound; we bind E so the editor is discoverable.
- `test/artifact`: F3 opens the editor against a live catalog.

## Save format
New games still start from vanilla `itemdata.lbx`. Edits persist **in the
current campaign save only**.

Old saves stored remaining premade items as a name list:

```json
"artifact-pool": ["Pummel Mace", "..."]
```

On load those names were looked up in a fresh `ReadArtifacts()` from
`itemdata.lbx`, so an in-memory edit could not survive save/load.

New saves still write that name list (so older builds can load the file
and reconstruct availability from vanilla `itemdata.lbx`), and add:

```json
"artifact-catalog": [ { "name": "...", "type": ..., "powers": [...], ... } ],
"artifact-available": [true, false, ...]
```

Load prefers `artifact-catalog` + `artifact-available` when the catalog
array is non-empty. Old saves without those fields keep working via the
name list + vanilla `itemdata.lbx`.

`SerializedArtifact` also gains `catalog-index` so a slot keeps its
ITEMDATA index across the round trip.

## What this does not do
- **Does not write `itemdata.lbx`.** ITEMMAKE and the Community Patch
  in-game editor both wrote the data file, so the next *new* game would
  see the edited catalog. We have no LBX writer, WASM cannot usefully
  write the copyrighted data files, and this remake should not mutate
  `data/mom/*.LBX`. A sidecar override for cross-game persistence would
  be a follow-up.
- Does not change items already sitting on heroes or in the vault.
- Does not start Enchant Item / Create Artifact, and does not reproduce
  the official ITEMMAKE book-requirement bug.

## Notes
- Headless tests in `artifact/catalog_test.go` cover award vs replace,
  old-save name lists, serialize round-trip of an edited unused slot, and
  realm-correct requirements.
- Verified: `go test ./...`, native `go build ./game/magic`, and
  `GOOS=js GOARCH=wasm go build` all pass.

Open as draft for manual in-game review (bind is E; or `go run ./test/artifact` and press F3).

### DIFF
--- a/game/magic/artifact/catalog.go
+++ b/game/magic/artifact/catalog.go
@@ -144,8 +144,8 @@ func (catalog *Catalog) Replace(index int, item *Artifact) {
     slot.Image = item.Image
     slot.Name = item.Name
     slot.Cost = item.Cost
-    slot.Powers = append([]Power{}, item.Powers...)
-    slot.Requirements = append([]Requirement{}, item.Requirements...)
+    slot.Powers = slices.Clone(item.Powers)
+    slot.Requirements = slices.Clone(item.Requirements)
     slot.CatalogIndex = index
 }
 
@@ -212,7 +212,7 @@ func (catalog *Catalog) AvailableMask() []bool {
         return nil
     }
 
-    return append([]bool{}, catalog.available...)
+    return slices.Clone(catalog.available)
 }
 
 func ReconstructCatalog(items []SerializedArtifact, available []bool, allSpells spellbook.Spells) *Catalog {
@@ -243,8 +243,8 @@ func CloneArtifact(src *Artifact) *Artifact {
     }
 
     clone := *src
-    clone.Powers = append([]Power{}, src.Powers...)
-    clone.Requirements = append([]Requirement{}, src.Requirements...)
+    clone.Powers = slices.Clone(src.Powers)
+    clone.Requirements = slices.Clone(src.Requirements)
     return &clone
 }
 
--- a/game/magic/artifact/catalog.go
+++ b/game/magic/artifact/catalog.go
@@ -1,0 +1,387 @@
+package artifact
+
+import (
+    "slices"
+
+    "github.com/kazzmir/master-of-magic/game/magic/data"
+    "github.com/kazzmir/master-of-magic/game/magic/spellbook"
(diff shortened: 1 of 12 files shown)


kazzmir:
Looks good. Two things:
1. if a hero already has item X and then you edit item X, the hero still has the old version. This is because Replace() clones the new artifact, but leaves the old one untouched. I could go either way on this, but when I initially tried it I expected the hero's weapon to just be updated.
2. Its a little tedious to get to item 150 in the item editor, since you have to press the right arrow 150 times. It might be nice if there was some sort of scroll list where you could see all the items and pick one fairly quickly. That can be a follow up pr

for #1, was it your intention that existing artifacts not get mutated?

gabe647:
1. It was intentional (ITEMMAKE: already-awarded copies stay as they were). For an in-game editor — especially WASM, where we don't write `itemdata.lbx` — live update makes more sense.

`Replace` now copies fields into the existing slot pointer instead of cloning. Heroes/vaults that already hold that item see the edit. Availability is unchanged. Pushed in 3f2f875.

2. Agreed, follow-up. A pick-from-list UI rather than 150 arrow clicks.

kazzmir (approved):

PR #716, opened 2026-06-03
ai: lair-clearing and goal-driven planning for Enemy2 AI

alexfibreplanner:
## Summary
Gives the computer wizards (Enemy2AI) a layered, weighted goal system plus the engine plumbing required to act on it. Non-brown AI wizards now use Enemy2AI.

## What it does
- **Goal planning:** `ComputeGoals` builds a weighted goal list — defeat enemies, build cities, increase power, defend cities, connect cities with roads, meld magic nodes, enchant units, planar travel — with sub-goals.
- **Lair clearing:** catalogues nearby lairs/nodes, estimates the attack strength needed (with a safety margin), and only commits armies strong enough to clear them.
- **Road building:** `AIBuildRoadDecision` + engine handling that walks engineers along a path and resumes in-progress road work each turn.
- **Node melding:** `AIMeldNodeDecision` + `tryMeldNode`, which melds the node a magic spirit is standing on once it arrives.
- **Per-unit enchantments:** `AICastUnitSpellDecision` lets the AI cast unit spells on a chosen target (e.g. Water Walking on scouts, combat buffs on elite stacks).
- **Planar travel:** `AIPlaneShiftDecision` and tower ferrying so a matured AI can expand to the other plane.
- **Settler pacing:** caps concurrent settlers so cities don't all spam settlers at once.
- **Engine fixes:** `DoEncounter` now returns the combat state so a fleeing AI attacker is moved back to its origin tile; ship-stack pathing fix; lair treasure-budget rebalance; added `Map.GetMagicNodeLocations` / `GetOpenTowerLocations`.

## Test plan
- [x] `go build ./game/magic`
- [x] `go vet ./game/magic/ai/ ./game/magic/game/ ./game/magic/player/ ./game/magic/maplib/`
- [x] `go test ./game/magic/ai/`
- [ ] Watch an AI-vs-AI game (`-watch`) and confirm AI clears lairs, builds roads, melds nodes, and expands across planes

PR #738, opened 2026-08-16
game: do not path a stack onto land if it contains a ship

gabe647:
## Summary
Split out of [#716](https://github.com/kazzmir/master-of-magic/pull/716) as requested in review.

AI movement (`doAiMoveUnit`) does not split inactive units off the stack
before pathing (there is a FIXME). `CanMoveOnLand(true)` only looks at
active units, so an inactive warship rides inland with the swordsmen —
the ship-on-land screenshot on #716.

`FindPath` and `ComputeTerrainCost` now call `CanMoveOnLand(false)` for
land destinations so **any** ship in the stack blocks a land move.

Human movement is unchanged: it already splits inactive units off before
`FindPath`.

## Tests
`path_test.go`:
- water tile, inactive Warship + active Swordsmen → no path onto land
- coast tile, same mix → no path further inland

`go test ./game/magic/game/ -run TestPathBasic` passes.

### DIFF
--- a/game/magic/game/path_test.go
+++ b/game/magic/game/path_test.go
@@ -199,4 +199,41 @@ func TestPathBasic(test *testing.T) {
     if !checkValidPathUnexplored(2, 1, units.HighMenSwordsmen) {
         test.Errorf("Land walker should be able to path to unexplored water tile")
     }
+
+    // an inactive ship sharing a stack with active land units must not be
+    // dragged onto land. AI movement does not split inactive units off first
+    // (human movement does), so pathfinding has to reject the move.
+    func() {
+        player1 := playerlib.MakePlayer(setup.WizardCustom{}, true, 3, 1, map[herolib.HeroType]string{}, &model)
+
+        warship := player1.AddUnit(units.MakeOverworldUnit(units.Warship, 1, 0, data.PlaneArcanus))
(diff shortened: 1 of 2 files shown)


game/magic/game/path_test.go
```diff
@@ -199,4 +199,41 @@ func TestPathBasic(test *testing.T) {
...
+            test.Errorf("stack with an inactive ship must not path from water onto land")
+        }
+    }()
+
+    // same rule once the mixed stack is already on the coast: active land
+    // units must not walk further inland while the ship is still in the stack
+    func() {
+        player1 := playerlib.MakePlayer(setup.WizardCustom{}, true, 3, 1, map[herolib.HeroType]string{}, &model)
+
+        warship := player1.AddUnit(units.MakeOverworldUnit(units.Warship, 2, 0, data.PlaneArcanus))
```
kazzmir: im not sure if this test makes sense. sailing units can be on ocean tiles or shore tiles, both are considered 'not land', so the warship could never be at 2,0 which is a land tile

gabe647:
Agreed — sailing units only sit on ocean/shore, both `!IsLand`, so a warship at 2,0 (land) was not a legal setup.

Dropped that second case. The remaining test is the real one: inactive warship + active swordsmen on ocean (1,0) must not path onto land (2,0).

kazzmir (approved):

PR #744, opened 2026-08-31
ai: wire difficulty and Aggressive AI into Enemy2

gabe647:
## Summary
- Enemy2 now uses `GetDifficulty()` plus a session-only **Aggressive AI** checkbox (next to Random Events). Classic Average is not wildly meaner; Aggressive and Hard+ settle earlier, will queue settlers at 0 food, and start roads at 2 cities.
- City offense no longer requires a visible rival **stack** first: wizards march on visible or remembered scouted cities. Settlers/melders stay off attacks; lair raid margins are unchanged.
- Tests cover the new gates without LBX (city attack without stacks, Classic vs Aggressive settler/food/road knobs).

## Watch the AI
Need original LBX files on disk (not required for unit tests). Then:

```bash
go build -o magic ./game/magic
./magic -data data/mom -watch -music=false
```

(`-data` can be any directory or zip of the `.LBX` files.) Watch mode is AI-only (4 opponents, Average). Year summaries print to the log (`Cities:`, `Food:`). Turn on **Aggressive AI** from the in-game settings overlay (same row as Random Events) to compare; speed slider is at the bottom of the watch UI, press D for goal debug.

## Test plan
- [ ] `go test ./game/magic/ai ./game/magic/settings` (no LBX)
- [ ] Settings: Aggressive AI defaults off; toggling it live affects Enemy2
- [ ] Watch Classic: second cities well before the old ~year 150 stall; food-0 empires still hold settlers
- [ ] Watch Aggressive: earlier settlers, 2-city roads, small stacks marching on seen hamlets with no rival stack on the map
- [ ] Confirm settlers/melders are not sent on city attacks

Made with [Cursor](https://cursor.com)

### DIFF
--- a/game/magic/game/game.go
+++ b/game/magic/game/game.go
@@ -628,6 +628,12 @@ func MakeGameWithModel(lbxCache *lbx.LbxCache, music *musiclib.Music, gameSettin
 
     // game.Model = MakeGameModel(terrainData, settings, data.PlaneArcanus, game.Events, heroNames, game.AllSpells(), createArtifactPool(lbxCache), buildingInfo)
     game.Model = makeModel(lbxCache, game.Events)
+    game.Model.IsAggressiveAI = func() bool {
+        if gameSettings == nil {
+            return false
+        }
+        return gameSettings.AggressiveAI
+    }
 
     game.HudUI = game.MakeHudUI()
     game.PushDrawer(func(screen *ebiten.Image){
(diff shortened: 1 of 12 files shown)


game/magic/game/game.go
```diff
@@ -628,6 +628,12 @@ func MakeGameWithModel(lbxCache *lbx.LbxCache, music *musiclib.Music, gameSettin
 
     // game.Model = MakeGameModel(terrainData, settings, data.PlaneArcanus, game.Events, heroNames, game.AllSpells(), createArtifactPool(lbxCache), buildingInfo)
     game.Model = makeModel(lbxCache, game.Events)
+    game.Model.IsAggressiveAI = func() bool {
```
kazzmir, line 631: this is ok for now but I think its kind of weird to set this one function dynamically. maybe model just needs a way to read the current settings, like a SettingsProvider interface or something

kazzmir (approved):

PR #746, opened 2026-09-08
camera: clamp overland view to the polar map edges

gabe647:
Give the overland camera the map height so `GetZoomedY` cannot scroll past the north/south ice. City-preview cameras with `MapHeight` 0 keep the old unclamped formula.

Split out of mixed local POC work on `test-today`.

## Changes
- `Camera.MapHeight` / `SetMapHeight`, used when a real overland map is attached
- Clamp `GetZoomedY` to `[0, height - visibleTilesY]`
- `doMoveCamera` centers with a simple Y clamp instead of the old zoom-loop nudge
- Headless camera tests for north edge, south edge, interior, and unset height

## Test plan
- [x] `go test ./game/magic/camera/...`
- In-game: zoom at the north and south poles and confirm no empty band past the ice


kazzmir:
I am able to get a bunch of blank tiles to show up depending on where I move the camera.
image

Steps:
1. Be at any zoom level
2. Move camera to bottom of map
3. Once camera is at the bottom, keep trying to move the camera further down. The camera won't move but blank tiles will start to appear at the top of the map.

The same thing happens if you are at the top of the map and keep trying to move up

kazzmir:
The bottom of the map seems fixed now, I cant get blank tiles to show up. At the top of the map I don't see the polar ice layer though.
image

In the minimap the ice layer is there, and the camera bounds look like the ice should be shown.

PR #749, opened 2026-09-08
terrain: paint a one-row tundra cap at each pole

gabe647:
Match original MoM's full-width polar ice and keep lairs and random magic nodes off those rows. Tiny test maps skip the cap so 3x3 fixtures stay land.

Split out of mixed local POC work on `test-today`.

## Changes
- `PolarIceRows` / `IsPolarIceRow` / `placePolarIce` after rivers, before `ResolveTiles`
- One tundra row at y=0 and y=height-1 on maps with at least 6 rows
- Encounter placement refuses polar ice rows
- Random magic nodes skip polar ice rows

## Test plan
- [x] `go test ./game/magic/terrain/...`
- New game: confirm a full-width tundra belt at both poles, no lairs on those rows

### DIFF
--- a/game/magic/terrain/map.go
+++ b/game/magic/terrain/map.go
@@ -352,6 +352,9 @@ func (map_ *Map) placeRandomTerrainTiles(plane data.Plane, continents []Continen
             }
 
             point := points[index]
+            if IsPolarIceRow(point.Y, map_.Rows()) {
+                continue
+            }
             if magicNodeOk(point) {
                 map_.Terrain[point.X][point.Y] = algorithm.ChooseRandomElement(magicTiles)
                 maxNodes -= 1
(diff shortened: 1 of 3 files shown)


game/magic/terrain/map.go
```diff
@@ -654,6 +657,41 @@ func (map_ *Map) SetTerrainAt(x int, y int, terrainType TerrainType, data *Terra
 }
 
 
+func PolarIceRows(rows int) int {
+    // original MoM had a single full-width tundra row at each pole. tiny
+    // maps skip it so 3x3 tile tests do not become all ice.
+    if rows < 6 {
```
kazzmir, line 663: presumably this is to avoid placing polar ice on maps used in test cases? ok for now but maybe a cleaner way is to pass in an extra parameter to the GenerateLand function that is a bool to allow polar ice, then the map object can have a field that stores the bool value for future reference

kazzmir (approved):

